PR #28, opened 2019-04-19
Feature/22

thirtytwobits:


src/pydsdlgen/version.py
```diff
@@ -3,6 +3,6 @@
 # This software is distributed under the terms of the MIT License.
 #
 
-__version__ = 0, 3, 0
```
pavel-kirienko: Are we officially stable now? I was just about to break the API of PyDSDL again (not because I'm a big fan of destruction but because I had a very cool idea of which I will write somewhere in a few days once I'm done with the serialization codegen). It shouldn't really affect the business end of *PyDSDLGen* (how do I spell it properly? we should have invented a better name like Yukon) but still, would you not agree that we're not yet completely certain about the stability of this library/tool? Would it not be wise to delay stabilization until we have libuavcan fully functional?

thirtytwobits: Yeah, I agree we're not fully stable. I guess I'm just reticent to put something out on Pypi that is going to break all the time. I'll add some guidance about its stability later today.

pavel-kirienko: > I'm just reticent to put something out on Pypi that is going to break all the time

I see what you did there. But the major version number of zero should be a sufficient indicator of the lack of stability guarantees.

Issue #64, opened 2019-09-05
Add support for implicit language filters


thirtytwobits:
Add an option to the Generator and to nnvg to populate the global namespace with a given language's filters. This will enable filters like `{{ foo | id }}` where the language for the `id` filter is implicit.

Issue #62, opened 2019-08-27
Allow library users to add template globals


pavel-kirienko:
Say, I want to add a new global. That doesn't seem to be possible right now save for accessing the private attribute directly.

pavel-kirienko:
Yeah, I already voiced this idea in #20 and then rejected it myself, but the reasons were different. Let's just have `additional_globals` in the constructor?

Issue #23, opened 2019-04-15
Stropping


pavel-kirienko:
DSDL definitions may contain identifiers which may not be allowed in some programming languages. This library should provide a way for the caller to specify which identifiers are not allowed. Upon encountering such identifiers, pydsdlgen would extend them as specified by the caller. For example:

* in Rust, the identifier would be prefixed with `r#` to form a raw identifier: `if` --> `r#if`
* in Python or C++ there is no concept of raw identifier; a common practice is to append an underscore: `if` --> `if_`. This, however, may lead to another name conflict if the source definition contains both `x_` and `x` for some x.
* Nim uses a pair of backticks: `if` --> `` `if` ``

Here, for example, is a list of reserved identifiers in Python 3.7:

```python
>>> import keyword, builtins
>>> list(sorted(keyword.kwlist + dir(builtins)))
['ArithmeticError', 'AssertionError', 'AttributeError', 'BaseException', 'BlockingIOError', 'BrokenPipeError',
'BufferError', 'BytesWarning', 'ChildProcessError', 'ConnectionAbortedError', 'ConnectionError',
'ConnectionRefusedError', 'ConnectionResetError', 'DeprecationWarning', 'EOFError', 'Ellipsis', 'EnvironmentError',
'Exception', 'False', 'False', 'FileExistsError', 'FileNotFoundError', 'FloatingPointError', 'FutureWarning',
'GeneratorExit', 'IOError', 'ImportError', 'ImportWarning', 'IndentationError', 'IndexError', 'InterruptedError',
'IsADirectoryError', 'KeyError', 'KeyboardInterrupt', 'LookupError', 'MemoryError', 'ModuleNotFoundError', 'NameError',
'None', 'None', 'NotADirectoryError', 'NotImplemented', 'NotImplementedError', 'OSError', 'OverflowError',
'PendingDeprecationWarning', 'PermissionError', 'ProcessLookupError', 'RecursionError', 'ReferenceError',
'ResourceWarning', 'RuntimeError', 'RuntimeWarning', 'StopAsyncIteration', 'StopIteration', 'SyntaxError',
'SyntaxWarning', 'SystemError', 'SystemExit', 'TabError', 'TimeoutError', 'True', 'True', 'TypeError',
'UnboundLocalError', 'UnicodeDecodeError', 'UnicodeEncodeError', 'UnicodeError', 'UnicodeTranslateError',
[...]

PR #65, opened 2019-09-05
Implementation of #23 Stropping and #62 Add globals

thirtytwobits:
This is a complete PR for the Stropping implementation and the library
access to add globals with some other work on #64 mixed in. It is not
a 1.1.3 release candidate until #64 is complete.


src/nunavut/jinja/lang/__init__.py
```diff
@@ -32,7 +32,7 @@ def get_supported_languages() -> typing.Iterable[str]:
     return __language_modules__.keys()
 
 
-def add_language_support(language_name: str, environment: Environment) -> None:
+def add_language_support(language_name: str, environment: Environment, make_implicit: bool = False) -> None:
```
pavel-kirienko: Missing docs for `make_implicit`

src/nunavut/jinja/lang/c.py
```diff
@@ -16,6 +16,153 @@
...
+    '_Alignof',
+    '_Atomic',
+    '_Bool',
+    '_Complex',
+    '_Generic',
+    '_Imaginary',
+    '_Noreturn',
+    '_Pragma',
+    '_Static_assert',
+    '_Thread_local'
```
pavel-kirienko: The following are all reserved:

- `_[A-Z_]+[a-zA-Z0-9_]*`
- `(is|to|str|mem|wcs|atomic_|cnd_|mtx_|thrd_|tss_|memory_)[a-z][a-zA-Z_0-9]*`
- `u?int[a-zA-Z_0-9]*_t`
- `E[A-Z0-9][a-zA-Z_0-9]*`
- `FE[A-Z][a-zA-Z_0-9]*`
- `U?INT[a-zA-Z_0-9]*_(MAX|MIN|C)`
- `(PRI|SCN)[a-zX][a-zA-Z_0-9]*`
- `LC_[A-Z][a-zA-Z_0-9]*`
- `SIG_?[A-Z][a-zA-Z_0-9]*`
- `TIME_[A-Z][a-zA-Z_0-9]*`
- `ATOMIC_[A-Z][a-zA-Z_0-9]*`

https://en.cppreference.com/w/c/language/identifier
https://en.cppreference.com/w/cpp/language/identifiers

Would it not be better to make this a list of regexps?

src/nunavut/jinja/lang/__init__.py
thirtytwobits: Oh, good catch. I also forgot to test for a double call to this.

src/nunavut/jinja/lang/c.py
thirtytwobits: > Would it not be better to make this a list of regexps?

It would now! Good catch. Thanks.

pavel-kirienko:
`_` as a prefix won't work for Python because by convention it affects the member visibility. In Python, underscores should be appended instead.

src/nunavut/jinja/lang/c.py
thirtytwobits: > (is|to|str|mem|wcs|atomic_|cnd_|mtx_|thrd_|tss_|memory_)[a-z][a-zA-Z_0-9]*

My reading of cppreference is that this pattern is actually `(is|to|str|mem|wcs|atomic_|cnd_|mtx_|thrd_|tss_|memory_)[a-z]+`. Am I missing something?

pavel-kirienko: It says "followed by a lowercase letter", not "by lowercase letters". Although, now that I'm thinking of it, it doesn't make much sense; perhaps they meant the latter but worded the description poorly (or I am simply misreading it). I tried looking at the standard but couldn't immediately find the relevant sections. Let's go with your interpretation.

thirtytwobits: I'll change to `^(is|to|str|mem|wcs|atomic_|cnd_|mtx_|thrd_|tss_|memory_|memory_order_)[a-z]{1}` which is more correct and probably more efficient.

PR #66, opened 2019-09-10
Milestone/0.1.3

thirtytwobits:
Resubmitting on a branch since I don't want to upload to Pypi until after I have all the 1.3 milestone issues fixed.


thirtytwobits:
@pavel-kirienko, if you could review and approve the final change to add the implicit support through nnvg then I can merge this to the milestone branch with will not push to Pypi (I made sure only master has the Pypi upload token). This will close #23, #62, and #64 .

src/nunavut/jinja/lang/py.py
```diff
@@ -53,3 +59,86 @@ def filter_to_template_unique_name(env: Environment, base_token: str) -> str:
...
+        {{ "I like python" | py.id }}
+        {{ "&because" | py.id }}
+        {{ "if" | py.id }}
+        {{ "if" | py.id('stropped_') }}
+
+    Results Example::
+
+       I_like_python
+       ZX0038because
+       _if
```
pavel-kirienko:
```suggestion
       if_
```

The code and the tests have been updated but not the docs.

pavel-kirienko:
A quick glance at the test did not reveal the answer so maybe I could just ask here: what would happen if I made a DSDL namespace named `pass` and then generated a Python package from it?

thirtytwobits:
> A quick glance at the test suite did not reveal the answer so maybe I could just ask here: what would happen if I made a DSDL namespace named `pass` and then generated a Python package from it?

Still working on this. Turned out this exposed a layering violation in Nunavut where I was only creating the concept of language support under the jinja package. I'm hoisting that up and factoring out the jinja adapters from the generic language support.

thirtytwobits:
Okay, I think theres a big refactor needed to absorb how we've changed language support in Nunavut. I'm thinking of pulling some of the globals into a an object called `LanguageContext` and creating decorated versions of Generator and Namespace from this object. Something like this:

image

Thoughts?

pavel-kirienko:
not bad

pavel-kirienko:
Should I review this now?

thirtytwobits:
Okay. This (finally) fixes the stropping/encoding to work on the namespace paths. It also is setup for deeper language integration going forward. Sorry for the churn. Please review. Thanks.

pavel-kirienko:
Okay. Could you have a look at my earlier comments, please?

src/nunavut/jinja/lang/c.py
```diff
@@ -16,6 +16,179 @@
...
+    re.compile(r'^(is|to|str|mem|wcs|atomic_|cnd_|mtx_|thrd_|tss_|memory_|memory_order_)[a-z]{1}'),
+    re.compile(r'^u?int[a-zA-Z_0-9]*_t'),
+    re.compile(r'^E[A-Z0-9]+'),
+    re.compile(r'^FE_[A-Z]{1}'),
+    re.compile(r'^U?INT[a-zA-Z_0-9]*_(MAX|MIN|C)'),
+    re.compile(r'^(PRI|SCN)[a-zX]{1}'),
+    re.compile(r'^LC_[A-Z]{1}'),
+    re.compile(r'^SIG_?[A-Z]{1}'),
+    re.compile(r'^TIME_[A-Z]{1}'),
+    re.compile(r'^ATOMIC_[A-Z]{1}')
```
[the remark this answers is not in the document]
thirtytwobits: sure. Why not.

src/nunavut/lang/py.py
```diff
@@ -0,0 +1,144 @@
...
+        {{ "I like python" | py.id }}
+        {{ "&because" | py.id }}
+        {{ "if" | py.id }}
+        {{ "if" | py.id('stropped_') }}
+
+    Results Example::
+
+       I_like_python
+       ZX0038because
+       _if
```
pavel-kirienko:
```suggestion
       if_
```

src/nunavut/lang/py.py
```diff
@@ -0,0 +1,144 @@
...
+       stropped_if
```
pavel-kirienko:
```suggestion
       ifstropped_
```
(this is probably not what you *meant* but it follows from the implementation)

Issue #63, opened 2019-08-27
Consider automatically constructing instance tests for all descendants of pydsdl.SerializableType


pavel-kirienko:
Like this (copy-paste from PyUAVCAN code generator):

```python
def _construct_instance_tests_from_root(root: typing.Type[object]) \
        -> typing.Dict[str, typing.Callable[[typing.Any], bool]]:
    out = {root.__name__: lambda x: isinstance(x, root)}
    for derived in root.__subclasses__():
        out.update(_construct_instance_tests_from_root(derived))
    return out

tests = _construct_instance_tests_from_root(pydsdl.SerializableType)

generator = nunavut.jinja.Generator(...,
                                    additional_tests=tests)
```

Use in templates like `{%- if t.element_type is BooleanType %}`.

PR #68, opened 2019-09-25
Issue #63 automatically creating type tests.

thirtytwobits:
Also adding the .vscode stuff which I forgot to do before.


.vscode/launch.json
```diff
@@ -0,0 +1,38 @@
...
+                     "--outdir", "${workspaceFolder}/.vscode/nunavut_out",
+                     "-l", "c",
+                     "${workspaceFolder}/test/gentest_dsdl/dsdl/uavcan"]
+        },
+        {
+            "name": "Pytest: current test",
+            "type": "python",
+            "request": "launch",
+            "module": "pytest",
+            "args": ["-k", "test_namespace_stropping"],
```
pavel-kirienko: Why is there a command specific to a particular test?

thirtytwobits: crap. Good catch.

Issue #60, opened 2019-07-27
Test against Windows


pavel-kirienko:
[Certain projects](https://pyuavcan.readthedocs.io/) that depend on this fine library (not nnvg) are going to support Windows soon. My preliminary assessment indicates that the library is at least generally functional with no obvious issues.

This might be relevant: https://github.com/UAVCAN/pydsdl/pull/28

thirtytwobits:
Making progress. Almost have it working.

thirtytwobits:
Fixed!

PR #71, opened 2019-09-30
Consolidating on buildkite

thirtytwobits:
Trying to reduce the number of build systems I have to deal with. I'm also looking to switch to sonarqube in the future so having everything on buildkite will make that easier.


pavel-kirienko:
How compatible is buildkite with #60?

thirtytwobits:
> How compatible is buildkite with #60?

Should be fine (https://buildkite.com/docs/agent/v3/windows) but there is a question of whose windows box we can use. I can bring one up in EC2 I suppose but I'll have to build that image myself.

pavel-kirienko:
Okay. They have stock Windows VMs over there in AppVeyor though, although it probably doesn't agree with the consolidation effort.

Issue #78, opened 2019-10-15
tox -p build fails sometimes


thirtytwobits:
we share resources between tox build platforms in the build/test_output folder that can cause concurrent access failures. Separate out test_output by platform and then renable -p auto for tox builds.

Issue #80, opened 2019-10-18
Use setuptool entrypoints properly


thirtytwobits:
I stupidly created nnvg as a file instead of using automatic script creation. Do this instead:

https://setuptools.readthedocs.io/en/latest/setuptools.html#automatic-script-creation

PR #81, opened 2019-10-18
Issue #80 and #78

thirtytwobits:
1. Much needed cleanup after last set of language abstraction changes.
2. Adding Sybil and fixing issue #80 
3. Doing pytest correctly which also fixes issue #78


thirtytwobits:
There are a few interesting changes here so you should take a look.

src/nunavut/lang/__init__.py
```diff
@@ -41,22 +57,50 @@ def get_module(self) -> typing.Any:
...
     """
 
-    _default_output_extensions = {
-        'c': '.h',
-        'cpp': '.hpp',
-        'js': '.js',
-        'py': '.py'
+    _language_defaults = {
+        'c': ('.h', '__module__'),
+        'cpp': ('.hpp', '__module__'),
```
pavel-kirienko: Why double underscores? Would `_module_` not be sufficient, considering that `_.*_` is a reserved pattern in DSDL? Double underscores scare me because they are *so long*, whenever you encounter one it just feels like the underscore is going to extend into infinity and never end, your life and whole existence would just become one epic underscore that can only be ended by the heat death of the universe :scream_cat:

thirtytwobits: Wow. That's terrifying. I'd better fix this.

thirtytwobits:
> You seem to be a big fan of tox.

Indeed I am. Besides the ability to repeat tests across several python versions it also replaces the need for Make or CMake to automate the build process. Keeps everything nice and Pythonic.

src/nunavut/lang/__init__.py
thirtytwobits: Hmm. __init__ is mandated by Python so that's a required choice. What about the rest. Any suggestions for what we should call C, C++, and JavaScript source that defines a DSDL namespace?

PR #86, opened 2019-10-22
Feature/67

thirtytwobits:
Let me know if this will unblock you.


pavel-kirienko:
`full_reference_name`/`short_reference_name` work well, but `to_template_unique_name` still seems to be malfunctioning here.

thirtytwobits:
> `full_reference_name`/`short_reference_name` work well, but `to_template_unique_name` still seems to be malfunctioning here.

That would make sense since this is not "identifier aware". I'll change this to be `to_template_unique_id` and apply stropping and encoding rules to the token. I'll also add a "scoping delimiter" that will allow for namespacing. So:

```
def filter_to_template_unique_id(base_identifier: str, scoping_delimiter: typing.Optional[str] = None, stropping: bool = True) -> str:
```
```
{{ "any.str.Foo" | to_template_unique_id(scoping_delimiter='.') }}
```
```
any_.str_._Foo0_
```

Does this make sense?

pavel-kirienko:
No no no, what I am saying is that the counter that generates the variable part of the name is broken. When you request an ID with the key `f` three times, you get `_f0_`, `_f0_`, `_f0_` instead of  `_f0_`, `_f1_`, `_f2_`. This is not related to stropping & stuff at all.

pavel-kirienko:
I don't expect key stropping to be useful.

thirtytwobits:
> f three times, you get _f0_, _f0_, _f0_ 

Huh. I have tests that cover this that are passing. Can you point me at an example?

Also, Can we log that as a separate issue so we can clear the backlog of changes behind the identifier stropping issues?

pavel-kirienko:
This is reproducible on the branch `nunavut-updates`, Nunavut is invoked here:

https://github.com/UAVCAN/pyuavcan/blob/6f1544763d7c1b1baa688d1cf63d35c3818ef53e/pyuavcan/dsdl/_compiler.py#L156-L192

Tomorrow I can extract that into an MWE if you want me to. The CI build is currently broken because Nunavut v0.1.9 the branch depends on is not yet released.

> Can we log that as a separate issue so we can clear the backlog of changes behind the identifier stropping issues?

Sure. The issue is not observable with Nunavut v0.1.2 (PyUAVCAN master works).

thirtytwobits:
> This is reproducible on the branch nunavut-updates, Nunavut is invoked here:

It'd help if you could point me to where in the templates you are seeing this. You do know that the uniqueness is per-type, not global to the session?

> Sure

Can you click 'approve' on my two PRs then?

pavel-kirienko:
> It'd help if you could point me to where in the templates you are seeing this.

Here, for example:

https://github.com/UAVCAN/pyuavcan/blob/6f1544763d7c1b1baa688d1cf63d35c3818ef53e/pyuavcan/dsdl/_templates/deserialization.j2#L83

I did not attempt yet to identify the exact scope where the issue can be observed.

> You do know that the uniqueness is per-type, not global to the session?

Yes. See this example of misgenerated code I posted here: https://github.com/UAVCAN/nunavut/pull/85#issuecomment-544948779

> Can you click 'approve' on my two PRs then?

Sorry, häd a stack overflow :/

thirtytwobits:
Added #88

Issue #88, opened 2019-10-23
to_template_unique_name is not working as expected


thirtytwobits:
The filter to_template_unique_name is malfunctioning here. Below is an example of misgenerated code where unique identifiers are not unique:

```
    @staticmethod
    def _deserialize_aligned_(_des_: Heartbeat_1_0._DeserializerTypeVar_) -> Heartbeat_1_0:
        assert _des_.consumed_bit_length % 8 == 0, 'Deserializer is not byte-aligned'
        _bit_length_base_ = _des_.consumed_bit_length
        _des_.require_remaining_bit_length(56)
        # BEGIN COMPOSITE DESERIALIZATION: uavcan.node.Heartbeat.1.0
        # BEGIN STRUCTURE FIELD DESERIALIZATION: saturated uint32 uptime
        # The temporary _f0_ holds the value of the field "uptime"
        _des_.require_remaining_bit_length(32)
        assert _des_.consumed_bit_length % 8 == 0, 'saturated uint32'
        _f0_ = _des_.fetch_aligned_u32()
        # END STRUCTURE FIELD DESERIALIZATION: saturated uint32 uptime
        # BEGIN STRUCTURE FIELD DESERIALIZATION: truncated uint2 health
        # The temporary _f0_ holds the value of the field "health"
        _des_.require_remaining_bit_length(2)
        assert _des_.consumed_bit_length % 8 == 0, 'truncated uint2'
        _f0_ = _des_.fetch_aligned_unsigned(2)
        # END STRUCTURE FIELD DESERIALIZATION: truncated uint2 health
        # BEGIN STRUCTURE FIELD DESERIALIZATION: truncated uint3 mode
        # The temporary _f0_ holds the value of the field "mode"
        _des_.require_remaining_bit_length(3)
        _f0_ = _des_.fetch_unaligned_unsigned(3)
        # END STRUCTURE FIELD DESERIALIZATION: truncated uint3 mode
        # BEGIN STRUCTURE FIELD DESERIALIZATION: truncated uint19 vendor_specific_status_code
        # The temporary _f0_ holds the value of the field "vendor_specific_status_code"
        _des_.require_remaining_bit_length(19)
        _f0_ = _des_.fetch_unaligned_unsigned(19)
        # END STRUCTURE FIELD DESERIALIZATION: truncated uint19 vendor_specific_status_code
        self = Heartbeat_1_0(uptime=_f0_,
                             health=_f0_,
                             mode=_f0_,
[...]

Issue #91, opened 2019-10-24
Generate C++


thirtytwobits:
That's it. Just generate C++. Easy right?

ASMfreaK:
The #236 PR added basic support for (de)serialization. More work should be done, of course, but current version is a quite bit useable. Maybe not yet on embedded uCs, but on bigger computers.

emrainey:
Can't wait for [variant support](https://en.cppreference.com/w/cpp/utility/variant) (C++17)

thirtytwobits:
variant support is already available.

PR #99, opened 2019-11-06
[WIP] Another stab at #88

pavel-kirienko:
Disabling the Jinja cache results in the unique name generator state disappearing from the global Jinja context:

    >       return _UniqueNameGenerator.ensure_generator_in_globals(env.globals)('c', adj_base_token, '_', '_')
    E       TypeError: 'NoneType' object is not callable
    .tox/py35-test/lib/python3.5/site-packages/nunavut/lang/c.py:527: TypeError

This issue affects all languages. The following test modification indicates that the state gets replaced with None:

    ------------------------- src/nunavut/lang/__init__.py -------------------------
    index bd6524c..135a29a 100644
    @@ -226,7 +226,7 @@ class _UniqueNameGenerator:
        def ensure_generator_in_globals(cls, environment_globals: typing.Dict[str, typing.Any]) -> '_UniqueNameGenerator':
            from .. import TypeLocalGlobalKey

    -        if TypeLocalGlobalKey not in environment_globals:
    +        if environment_globals.get(TypeLocalGlobalKey) is None:
                environment_globals[TypeLocalGlobalKey] = cls()
            return typing.cast('_UniqueNameGenerator', environment_globals[TypeLocalGlobalKey])

I don't quite understand what is happening and why it worked on my machine last week.

@thirtytwobits do you find the added test logic sensible?

### DIFF
--- a/src/nunavut/jinja/__init__.py
+++ b/src/nunavut/jinja/__init__.py
@@ -375,7 +375,8 @@ class Generator(nunavut.generators.AbstractGenerator):
                                 keep_trailing_newline=True,
                                 lstrip_blocks=lstrip_blocks,
                                 trim_blocks=trim_blocks,
-                                auto_reload=False)
+                                auto_reload=False,
+                                cache_size=0)
 
         self._add_language_support()
 
--- a/src/nunavut/version.py
+++ b/src/nunavut/version.py
@@ -3,6 +3,6 @@
 # This software is distributed under the terms of the MIT License.
 #
 
-__version__ = "0.1.12"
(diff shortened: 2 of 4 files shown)


thirtytwobits:
So I rebased this on my PR and we do get past the None error but it still doesn't work the way you expect. I am uncertain that macros will have anything that isn't an argument reevaluated. For that you'll want to create another template and include it.

thirtytwobits:
Closing as I merged this into #91 before tracking down the final problems with it.

PR #100, opened 2019-11-06
Progress on #91

thirtytwobits:
1) Fixing a problem with unique identifier generation (Issue #88)
2) Generating C++ data structures (Issue #91)

Still TODO for Issue #91 is serialization for types.


pavel-kirienko:
I predict that the test I wrote for #99 will fail here because you seem to have encountered the same problem I had there. The culprit is where you added the extra check for none (can't link, am on mobile).

The caching issue may appear to be fixed unless you have a test to validate it, I can't find such test here.

Also, this is a 100-get ❤

thirtytwobits:
Let me know if this works for you. I've provided an example of using a template include instead of a macro to avoid the issues you are seeing with the unique identifiers. This includes a rebase with your change and then a modification of it.

pavel-kirienko:
> I am uncertain that macros will have anything that isn't an argument reevaluated.

(quote from #99) Macros are allowed to have side effects. They work well in the current PyUAVCAN master where I keep the state in a template variable:

https://github.com/UAVCAN/pyuavcan/blob/7cc79786e212940c4777a8ffa741cc0b6709b17a/pyuavcan/dsdl/_templates/unique_reference.j2

For some reason, they break only when the global state is added from the side of the Python API.

Switching to includes instead of macros is worse than the current solution so it's probably best to keep it unchanged then.

I think we are misusing Jinja, it's getting in the way again.

pavel-kirienko:
Maybe we should instead keep the state somewhere inside Nunavut instead of Jinja's globals so that it's not destroyed when a macro is invoked?

thirtytwobits:
Okay. This is the best I can figure out. There may be a way to create a "runtime macro" extension but that's out of scope for this change.

pavel-kirienko:
Now the counter is not reset when a new data type starts being generated. The actual caching problem is not solved unless we use the more verbose macro invocation form. I suggest we go back to the old hack, somehow it seems less hacky that the new solution.

pavel-kirienko:
Going back to the old solution means removal of the unique ID filter from Nunavut,

thirtytwobits:
> Now the counter is not reset when a new data type starts being generated. The actual caching problem is not solved unless we use the more verbose macro invocation form. I suggest we go back to the old hack, somehow it seems less hacky that the new solution.

Oh, I can reintroduce resetting the counter for each type but why should that matter?

pavel-kirienko:
Because the output code for a type X should not be dependent on that of the type Y, otherwise it complicates tracking the changes, and in compiled languages it may lead to unnecessary recompilation.

thirtytwobits:
Found a fix!

Issue #94, opened 2019-10-28
Remove stropping and encoding prefix from filter args. Make these global options


thirtytwobits:
The 'ZX' and '_' prefix/suffix stropping/encoding arguments in filters are not fully plumbed through, would cause a lot of extra noise if they were, and aren't useful to have differ between filters. We should remove any filter arguments that specify these and provides them as part of the language context instead.

thirtytwobits:
Actually, remove "stropping:bool" as well and make this global.

thirtytwobits:
I need to check but I may have inadvertently fixed this issue already.

thirtytwobits:
Yep. This happened.

PR #101, opened 2019-11-26
Issue #94 and progress on Issue #91

thirtytwobits:
Major refactor of language context to derive filter options from
configuration. The ability to disable or modify things like stropping and
encoding were fundamentally broken before this change.

This work was done in support of adding named types as a concept to
languages. This will allow generation of serialization code that is more
configurable and will avoid fixing decisions like the use of size_t into
templates instead moving these decisions into user-overridable
configuration.

Also moved a bunch of stuff into tox.ini after reading a screed by the pytest maintainers about how broken it is to put stuff in Setup.cfg. Even so, I'm learning this amounts to a re-arranging of deck chairs given PEP-518. *sigh*. It seems, perhaps, someday Python will get project configuration right but that day isn't here just yet.


.vscode/cmake-kits.json
```diff
@@ -0,0 +1,16 @@
+[
+    {
+        "name": "native-gcc",
+        "toolchainFile": "${workspaceFolder}/verification/cpp/cmake/toolchains/gcc-native.cmake",
+        "cmakeSettings": {
+            "LIBUAVCAN_FLAG_SET": "../cmake/compiler_flag_sets/native.cmake"
```
pavel-kirienko: How did Libuavcan creep into this? I understand this is probably just carried over from your Libuavcan project setup, but if you want it kept in the public Nunavut repo then it's best to find a better name to avoid undesirable association between Nunavut and Libuavcan. If I understand this thread correctly, they should not be interdependent: https://forum.uavcan.org/t/library-independent-code-generation/589

src/nunavut/lang/cpp/templates/_composite_type.j2
```diff
@@ -8,4 +8,12 @@
     {{ field.data_type | declaration }} {{ field.name | id }};
 {%- endif -%}
 {%- endfor %}
+    static {{ typename_unsigned_length }} serialize(std::vector<{{ typename_byte }}>& inout_byte_buffer, {{ typename_unsigned_length }} bit_offset, {{ composite_type | declaration }}& t)
```
pavel-kirienko: The line is longer than 120 characters.

src/nunavut/templates.py
```diff
@@ -0,0 +1,120 @@
...
+# Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# Copyright (C) 2018-2019  UAVCAN Development Team  <uavcan.org>
+# This software is distributed under the terms of the MIT License.
+#
+"""
+Abstractions around template engine internals.
+"""
+import typing
+
+EnvironmentFilterAttributeName = 'environmentfilter'
```
pavel-kirienko: Per PEP8, this should be named in `SCREAMING_SNAKE_CASE`. Why is it different here?

src/nunavut/templates.py
```diff
@@ -0,0 +1,120 @@
...
+
+class SupportsTemplateContext:
+    """
+    Provided as a pseudo
+    `protocol <https://mypy.readthedocs.io/en/latest/protocols.html#simple-user-defined-protocols/>`_.
+    (in anticipation of that becoming part of the core Python typing someday).
+    """
+
+
+def templateEnvironmentFilter(filter_func: typing.Callable) -> typing.Callable[..., str]:
```
pavel-kirienko: Per PEP8, this should be named in `snake_case`. Why is it different here?

verification/cpp/include/nunavut/support.hpp
```diff
@@ -0,0 +1,63 @@
...
+ * @param  src              The byte array to copy from.
+ * @param  dst              The byte array to copy data into.
+ * @param  dst_offset_bits  The offset, in bits, from the start of the dst array to
+ *                          start writing to.
+ * @param  length_bits      The total length of bits to copy. The caller must ensure
+ *                          that the size of src and dst are >= this value.
+ *
+ * @return The number of bits copied.
+ */
+template <typename SIZE_TYPE, typename BYTE_TYPE>
```
pavel-kirienko: This should be `SizeType` and `ByteType`, per the coding conventions.

verification/cpp/suite/test_support.cpp
```diff
@@ -0,0 +1,50 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Tests of the Nunavut support header.
+ */
+#include "gmock/gmock.h"
+#include "nunavut/support.hpp"
+#include <vector>
+
+template <typename SIZE_TYPE_FOR_TEST, typename BYTE_TYPE_FOR_TEST>
```
pavel-kirienko: Should be CamelCase here, too.

pavel-kirienko:
> Also moved a bunch of stuff into tox.ini after reading a screed by the pytest maintainers about how broken it is to put stuff in Setup.cfg.

Could you share the link, please?

.vscode/cmake-kits.json
thirtytwobits: Cruft. I'll rename this. There is no libuavcan dependency.

src/nunavut/templates.py
thirtytwobits: REALLY? MY_PEP8_LINTER_DIDN'T_COMPLAIN. LET_ME_LOOK_INTO_THIS!

thirtytwobits:
> > Also moved a bunch of stuff into tox.ini after reading a screed by the pytest maintainers about how broken it is to put stuff in Setup.cfg.
> 
> Could you share the link, please?

https://github.com/pytest-dev/pytest/issues/3523

src/nunavut/templates.py
thirtytwobits: I believe decorators are in camel case for PEP8?

pavel-kirienko: PEP8 says this:

> mixedCase is allowed only in contexts where that's already the prevailing style (e.g. threading.py), to retain backwards compatibility.

Decorators are generally functions so they should be in snake_case.

src/nunavut/templates.py
pavel-kirienko: PEP8 says:

> Constants are usually defined on a module level and written in all capital letters with underscores separating words. Examples include MAX_OVERFLOW and TOTAL.

src/nunavut/templates.py
thirtytwobits: I'd like to keep this camelcase to be more consistent with JINJA decorators.

thirtytwobits: image

PR #107, opened 2020-04-08
PyDSDL v1.3+: Support TaggedUnionType

pavel-kirienko:
This changeset fixes the dependency list handling which was broken by https://github.com/UAVCAN/pydsdl/pull/34. This functionality was not covered by tests and I left it this way (sorry).

I also fixed the mismatch between the documented behavior and the actual behavior in `py.filter_imports()`: the list should contain unique entries.

The tests fail because the rest of the codebase is not yet updated to use TaggedUnionType. @thirtytwobits what should be done about this?

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -8,7 +8,7 @@ long_description = file: README.rst
 long_description_content_type = text/x-rst
 license = MIT
 license_file = LICENSE.rst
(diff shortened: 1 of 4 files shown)


pavel-kirienko:
Closing due to https://github.com/UAVCAN/pydsdl/pull/39.

PR #109, opened 2020-04-11
Fix Python dependency imports, fix minor PyDSDL API usage issues

### LINKED ISSUE #108: Do we care about code generation performance?
It is not something that is done often so the benefit of such optimizations is low but it is non-zero. Experimentation may require one to re-generate code multiple times in a short period of time; doing so incrementally is currently impossible because we can only operate at the root namespace level (there are reasons for that related to data type compatibility, long story).

I wouldn't raise this issue if not for the fact that running a benign command like `pyuavcan -vv dsdl-gen-pkg ~/uavcan/public_regulated_data_types/uavcan/` takes 55 seconds, of which 51 appear to be used by Jinja.

See the performance profiles (output ordered by cumulative and total time): [perf.tar.gz](https://github.com/UAVCAN/nunavut/files/4464859/perf.tar.gz)

Are we OK with that or is there a cheap way to make Jinja faster? The performance issues were probably introduced when we disabled the cache to implement predictable side effects necessary for unique identifier generation:

https://github.com/thirtytwobits/nunavut/blame/7fd2ef1640430f746c8d0c00347a04fa2920d966/src/nunavut/jinja/__init__.py#L410

pavel-kirienko:
Fixes #108 by enabling [bytecode caching](https://jinja.palletsprojects.com/en/2.11.x/api/#bytecode-cache) in Jinja. PyUAVCAN tests pass. Ready to merge if CI is green.

The case described in #108 is down from 55 seconds to 7.7 seconds. Performance profile: [out-bytecode-cache-cumulative.tar.gz](https://github.com/UAVCAN/nunavut/files/4464928/out-bytecode-cache-cumulative.tar.gz)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -8,7 +8,7 @@ long_description = file: README.rst
 long_description_content_type = text/x-rst
 license = MIT
 license_file = LICENSE.rst
-keywords = uavcan, dsdl, can, can-bus, codegen, 
+keywords = uavcan, dsdl, can, can-bus, codegen,
 classifiers =
     Development Status :: 3 - Alpha
     Environment :: Console
@@ -35,7 +35,7 @@ package_dir=
     =src
 packages=find:
 install_requires=
-    pydsdl ~= 1.2
+    pydsdl ~= 1.4
     setuptools
 
 zip_safe = False
--- a/src/nunavut/cli.py
+++ b/src/nunavut/cli.py
@@ -17,7 +17,7 @@ import textwrap
 import typing
 
 
-def _run(args: argparse.Namespace, extra_includes: str) -> int:  # noqa: C901
+def _run(args: argparse.Namespace, extra_includes: typing.List[str]) -> int:  # noqa: C901
     '''
         Post command-line setup and parsing logic to execute nunavut
         library routines based on input.
--- a/src/nunavut/jinja/__init__.py
+++ b/src/nunavut/jinja/__init__.py
@@ -25,7 +25,7 @@ import nunavut.postprocessors
 from nunavut.jinja.jinja2 import (ChoiceLoader, Environment, FileSystemLoader,
                                   PackageLoader, StrictUndefined, Template,
                                   TemplateAssertionError, nodes,
-                                  select_autoescape)
+                                  select_autoescape, FileSystemBytecodeCache)
 from nunavut.jinja.jinja2.ext import Extension
 from nunavut.jinja.jinja2.parser import Parser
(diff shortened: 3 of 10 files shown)


pavel-kirienko:
With this change, the PyUAVCAN test suite is executed in 14 minutes vs. 30 minutes.

pavel-kirienko:
At first glance, the build failures seem unrelated to my changes but I didn't dig yet.

pavel-kirienko:
Yeah, bytecode caching breaks the tests. I tried adding extra sharding to prevent false cache hits but a few quick attempts to resolve it did not work.

@thirtytwobits I am going to need your advice here -- do we want to pursue this? Jinja is slow enough for it to be a mild problem (we can ignore it but if there is a cheap way to fix it we should fix it).

thirtytwobits:
> Yeah, bytecode caching breaks the tests. I tried adding extra sharding to prevent false cache hits but a few quick attempts to resolve it did not work.
> 
> @thirtytwobits I am going to need your advice here -- do we want to pursue this? Jinja is slow enough for it to be a mild problem (we can ignore it but if there is a cheap way to fix it we should fix it).

I don't think this is going to ever work since we never have a template that results in the same output. The single biggest performance problem is well-documented; don't use line post-processors (e.g. https://nunavut.readthedocs.io/en/latest/docs/api/nunavut.html#nunavut.postprocessors.TrimTrailingWhitespace). Enabling any line post-processors will force Nunavut to create an intermediate line-buffer that requires it to rebuffer everything.

pavel-kirienko:
> I don't think this is going to ever work since we never have a template that results in the same output.

Observe that this is not template caching I am talking about here, but *bytecode caching*. It works with PyUAVCAN (according to its generated code validation suite) but breaks with Nunavut's tests; I didn't investigate why exactly that happens.

> The single biggest performance problem is well-documented; don't use line post-processors (e.g. https://nunavut.readthedocs.io/en/latest/docs/api/nunavut.html#nunavut.postprocessors.TrimTrailingWhitespace). Enabling any line post-processors will force Nunavut to create an intermediate line-buffer that requires it to rebuffer everything.

The negative effect of line post-processing is negligible.

Compared to the original case reported in #108, removal of line post-processing shaves off 2 seconds, reducing the total run time from 55 to 53 seconds: [no-line-post-processors.tar.gz](https://github.com/UAVCAN/nunavut/files/4471300/no-line-post-processors.tar.gz)

In the profile, you can see that most of the work is related to parsing Jinja templates, apparently.

pavel-kirienko:
I have removed bytecode caching from this PR. Let's focus on the other changes for now.

PR #110, opened 2020-05-10
Progress on C++ type generation

thirtytwobits:
1. reorganization of code to add "support library" concept
2. adding built-in C++ templates

TODO:
1. finish and fully test the support library
2. generate serialization routines as part of the types.


src/nunavut/lang/cpp/templates/_fields_as_union.j2
```diff
@@ -0,0 +1,13 @@
+    {{ composite_type.tag_field_type | declaration }} tag;
```
pavel-kirienko, line 1:
```suggestion
    {{ composite_type.tag_field_type | declaration }} _tag_;
```

Breaks if the union contains an attribute named `tag`.

src/nunavut/lang/cpp/templates/_fields_as_union.j2
```diff
@@ -0,0 +1,13 @@
...
+    union {
+{%- for field in composite_type.fields -%}
+{%- if field is not padding %}
+        {{ field.data_type | declaration }} {{ field.name | id }};
+{%- endif -%}
+{%- endfor %}
+    };
+{%- for field in composite_type.fields -%}
+{%- if field is not padding %}
+    bool {{ ('is_' + field.name)  | id }}() { return (tag == {{ loop.index0 }}); }
```
pavel-kirienko, line 11:
```suggestion
    bool {{ ('is_' + field.name)  | id }}() const { return (_tag_ == {{ loop.index0 }}); }
```

src/nunavut/lang/cpp/templates/_fields_as_union.j2
```diff
@@ -0,0 +1,13 @@
...
+{%- endif -%}
+{%- endfor %}
```
pavel-kirienko, line 13: If we chose to go this way, then we should hide the inner union and the tag from the user completely by making them private. Exposing the aliased memory is unsafe and MISRA linters will be complaining about that in the user code. We need getters (const and not) that return a pointer to the member if the tag is correct and nullptr otherwise (they will also obviate `is_...()`). A setter per field is also needed to allow switching the tag value.

The above is undesirable though. Ideally, we should avoid reinventing this little wheel completely and just use `std::variant` as all normal people do, but it will cause issues with C++14/11. To avoid said issues, we could use `std::variant` if the detected version is C++17 or newer, otherwise, fallback to [`bpstd::variant`](https://github.com/bitwizeshift/BackportCpp/blob/master/include/bpstd/variant.hpp) or some other alternative. Handling raw unions without the typesafe infrastructure that is readily available for `std::variant` will be hard, error-prone, and require much manual boilerplating.

src/nunavut/lang/cpp/support/serialization.hpp
```diff
@@ -1,73 +1,318 @@
...
+union Fp32
+{
+    std::uint32_t u;
+    float         f;
+};
+
+}  // namespace internal
+
+template <typename NativeType>
+NativeType nativeIeeeToHalf(float value)
```
pavel-kirienko, line 73:
```suggestion
NativeType nativeIeeeToHalf(const float value)
```

src/nunavut/lang/cpp/support/serialization.hpp
```diff
@@ -1,73 +1,318 @@
...
+        out = 0xFFFF & static_cast<NativeType>(in.u >> 13); /* Take the bits! */
+    }
+
+    out = 0xFFFF & static_cast<NativeType>(out | (sign >> 16));
+
+    return out;
+}
+
+template <typename NativeType>
+float halfToNativeIeee(NativeType value)
```
pavel-kirienko, line 121:
```suggestion
float halfToNativeIeee(const NativeType value)
```

src/nunavut/lang/cpp/support/serialization.hpp
```diff
@@ -1,73 +1,318 @@
...
+        out.u |= 255U << 23;
+    }
+    out.u |= (value & 0x8000U) << 16; /* sign bit */
+
+    return out.f;
+}
+
+template <unsigned BitLen>
+static typename NativeFloatSelector<BitLen>::int_storage_type toIeee(
+    typename NativeFloatSelector<BitLen>::float_type value)
```
pavel-kirienko, line 148:
```suggestion
    const typename NativeFloatSelector<BitLen>::float_type value)
```

src/nunavut/lang/cpp/support/serialization.hpp
```diff
@@ -1,73 +1,318 @@
...
+        typename NativeFloatSelector<BitLen>::float_type       f;
+    } u;
+    static_assert(sizeof(u.f) * 8 == BitLen, "Float type selected does not match BitLen");
+    u.f = value;
+    return u.i;
 }
 
+template <unsigned BitLen>
+static typename NativeFloatSelector<BitLen>::float_type toNative(
+    typename NativeFloatSelector<BitLen>::int_storage_type value)
```
pavel-kirienko, line 163:
```suggestion
    const typename NativeFloatSelector<BitLen>::int_storage_type value)
```

src/nunavut/lang/cpp/support/serialization.hpp
```diff
@@ -1,73 +1,318 @@
...
+        typename NativeFloatSelector<BitLen>::Type             f;
+    } u;
+    static_assert(sizeof(u.f) * 8 == BitLen, "Float type selected does not match BitLen");
+    u.i = value;
+    return u.f;
+}
+
+}  // namespace IEEE754Converter
+
+template <typename ByteType, std::size_t static_capacity_bits>
```
pavel-kirienko, line 178: Why do we need to parametrize the byte type?

src/nunavut/lang/cpp/support/serialization.hpp
```diff
@@ -1,73 +1,318 @@
...
+class LittleEndianSerializer
+{
+public:
+    static_assert(!std::is_signed<ByteType>::value, "ByteType must be unsigned");
+    static_assert(std::is_integral<ByteType>::value, "ByteType must be an integer");
+
+    LittleEndianSerializer(std::vector<ByteType>& inout_buffer, std::size_t bit_offset)
```
pavel-kirienko, line 185: We should avoid reliance on the heap unless it introduces new application constraints or it is expensive to do so. Here, neither serializer nor deserializer need dynamic memory because the capacity is statically known, so the vector should be replaced with a pointer to a static external buffer, and the class should be equipped with a getter that can be queried upon completion of the serialization to determine the amount of the buffer space used.

src/nunavut/lang/cpp/support/serialization.hpp
```diff
@@ -1,73 +1,318 @@
...
+    } u;
+    static_assert(sizeof(u.f) * 8 == BitLen, "Float type selected does not match BitLen");
+    u.i = value;
+    return u.f;
+}
+
+}  // namespace IEEE754Converter
+
+template <typename ByteType, std::size_t static_capacity_bits>
+class LittleEndianSerializer
```
pavel-kirienko, line 179: There are numerous violations of the naming conventions in this class.

pavel-kirienko: There are numerous MISRA violations in this class.

src/nunavut/lang/cpp/support/serialization.hpp
pavel-kirienko: May I suggest that we introduce macros for the language features that are not available in older versions of C++ to make better use of the newer versions? Particularly, I would like to use `[[nodiscard]]` via some `NUNAVUT_NODISCARD` which evaluates to nothing in C++11/14.

pavel-kirienko:
Are we going to have a heapless array class in the support library that preallocates the storage (like in v0), or are we going to rely on the STL containers instead (with a deterministic heap allocator)? If it's the latter, how are we going to handle OOM given that exception handling may be disabled?

I am inclined to favor the heapless approach.

thirtytwobits:
> Are we going to have a heapless array class in the support library that preallocates the storage (like in v0), or are we going to rely on the STL containers instead (with a deterministic heap allocator)? If it's the latter, how are we going to handle OOM given that exception handling may be disabled?
> 
> I am inclined to favor the heapless approach.

I have a grand hope that we can use vector with a deterministic allocator provided. I haven't evaluated vectors behavior in the face of OOM with `-fno-exceptions`. I'll look into that when I get back to C++-land.

src/nunavut/lang/cpp/support/serialization.hpp
thirtytwobits: First off, MISRA doesn't cover C++11 and newer so it's not a good standard to reference. Fortunately they agree: https://www.misra.org.uk/LinkClick.aspx?fileticket=n7HwIG-0JYk%3D&tabid=59

Now, as for Autosar, yeah, this is a work-in-progress. Seeing how many comments you have I'm inclined to simply remove this for now and come back to it later.

pavel-kirienko: I am referring to that subset of MISRA which is valid for C++11.

pavel-kirienko:
> I have a grand hope that we can use vector with a deterministic allocator provided.

What benefits are there compared to the heapless approach?

If the heap-based approach is chosen, the API of the (de-)serializer will need to be updated to permit in-place serialization where the class is provided with a particular slice of the buffer instead of a dedicated new buffer. This is needed to properly support nested types.

src/nunavut/lang/cpp/support/serialization.hpp
thirtytwobits: Going to delete all this for now.

src/nunavut/lang/cpp/support/serialization.hpp
thirtytwobits: Going to delete all this for now.

src/nunavut/lang/cpp/support/serialization.hpp
thirtytwobits: Going to delete all this for now.

src/nunavut/lang/cpp/support/serialization.hpp
thirtytwobits: Going to delete all this for now.

src/nunavut/lang/cpp/support/serialization.hpp
thirtytwobits: Going to delete all this for now.

src/nunavut/lang/cpp/support/serialization.hpp
thirtytwobits: Going to delete all this for now.

src/nunavut/lang/cpp/support/serialization.hpp
thirtytwobits: Going to delete all this for now.

src/nunavut/lang/cpp/templates/_fields_as_union.j2
thirtytwobits: I'm going to comment this out so we can get this right in the future but not expose the current implementation to the public (without warning). New change incoming...

thirtytwobits:
> > I have a grand hope that we can use vector with a deterministic allocator provided.
> 
> What benefits are there compared to the heapless approach?
> 
> If the heap-based approach is chosen, the API of the (de-)serializer will need to be updated to permit in-place serialization where the class is provided with a particular slice of the buffer instead of a dedicated new buffer. This is needed to properly support nested types.

My desire is to separate memory management from serialization such that the consumer of the generated code can provide memory allocators appropriate to their platform. I believe the specifics of this discussion are aimed at libuavcan and not the generic C++ we want to emit from Nunavut. For libuavcan we will want to provide a deterministic and constant-complexity allocator. That said, even in libuavcan, we might want to allow linux or other high-performance, low-reliability systems to use something else including the default binary heap from their standard C++ library.

thirtytwobits:
Okay. Going to merge but I think the CI will break because I'm also streamlining the pipelines and this is the first run that will utilize my new configuration. I'll work on fixing this up as they break.

PR #113, opened 2020-05-26
Adding C PoD type generation

thirtytwobits:
This commit adds the ability to generate Plain-Ol-Datatypes a C from
internal templates. These templates include initializer methods and
array helper macros. The types are also forward compatible with C++.

To see an example of the template output do:
```
pytest -k test_realgen[c-False] --keep-generated
```
and you'll see the results under the folder:
```
build/gentest_dsdl_test_realgen\[c-False\]/
```

For example: 
```c
/*
 *
 * UAVCAN data structure definition.                                                                          +-+ +-+
 *                                                                                                            | | | |
 * AUTOGENERATED, DO NOT EDIT.                                                                                \  -  /
 *                                                                                                              ---
 *                                                                                                               o
 * +------------------------------------------------------------------------------------------------------------------+
 *
 * Generator:
 * nunavut-0.2.4 (serialization was not enabled)
 *
 *
 * Source File:
 * /nunavut/submodules/public_regulated_data_types/uavcan/register/Value.1.0.uavcan
 *
 * Template:
 * UnionType.j2
 *
 * Generated at:  2020-05-26 18:42:56.149848 UTC
 * Is deprecated: no
 * Fixed port ID: None
 * Full name:     uavcan.register.Value
 * Version:       1.0
 *
 */

#ifndef UAVCAN__REGISTER_VALUE
#define UAVCAN__REGISTER_VALUE

#include <stdint.h>
#include "uavcan/primitive/Empty_1_0.h"
#include "uavcan/primitive/String_1_0.h"
#include "uavcan/primitive/Unstructured_1_0.h"
#include "uavcan/primitive/array/Bit_1_0.h"
#include "uavcan/primitive/array/Integer16_1_0.h"
#include "uavcan/primitive/array/Integer32_1_0.h"
#include "uavcan/primitive/array/Integer64_1_0.h"
#include "uavcan/primitive/array/Integer8_1_0.h"
#include "uavcan/primitive/array/Natural16_1_0.h"
#include "uavcan/primitive/array/Natural32_1_0.h"
#include "uavcan/primitive/array/Natural64_1_0.h"
#include "uavcan/primitive/array/Natural8_1_0.h"
#include "uavcan/primitive/array/Real16_1_0.h"
#include "uavcan/primitive/array/Real32_1_0.h"
#include "uavcan/primitive/array/Real64_1_0.h"

#ifdef __cplusplus
extern "C" {
#endif /* __cplusplus */

typedef struct
{    
    uint8_t _tag_;
    union {
        UAVCAN_PRIMITIVE_Empty_1_0 empty;
        UAVCAN_PRIMITIVE_String_1_0 _string;
        UAVCAN_PRIMITIVE_Unstructured_1_0 unstructured;
        UAVCAN_PRIMITIVE_ARRAY_Bit_1_0 bit;
        UAVCAN_PRIMITIVE_ARRAY_Integer64_1_0 integer64;
        UAVCAN_PRIMITIVE_ARRAY_Integer32_1_0 integer32;
        UAVCAN_PRIMITIVE_ARRAY_Integer16_1_0 integer16;
        UAVCAN_PRIMITIVE_ARRAY_Integer8_1_0 integer8;
        UAVCAN_PRIMITIVE_ARRAY_Natural64_1_0 natural64;
        UAVCAN_PRIMITIVE_ARRAY_Natural32_1_0 natural32;
        UAVCAN_PRIMITIVE_ARRAY_Natural16_1_0 natural16;
        UAVCAN_PRIMITIVE_ARRAY_Natural8_1_0 natural8;
        UAVCAN_PRIMITIVE_ARRAY_Real64_1_0 real64;
        UAVCAN_PRIMITIVE_ARRAY_Real32_1_0 real32;
        UAVCAN_PRIMITIVE_ARRAY_Real16_1_0 real16;
    };
} UAVCAN__REGISTER_Value_1_0;

/**
 * Initialize an UAVCAN__REGISTER_Value_1_0 instance to default values.
 * No memory is allocated within this method.
 *
 * @param inout_instance A structure instance to Initialize.
 * @return The inout_instance for fluency.
 */
inline UAVCAN__REGISTER_Value_1_0* init_uavcan__register_value_1_0(UAVCAN__REGISTER_Value_1_0* inout_instance)
{
    if (inout_instance)
    {
        inout_instance->_tag_ = 0;
        init_uavcan_primitive_empty_1_0(&(inout_instance->empty));
    }
    return inout_instance;
}
#define IS_UAVCAN__REGISTER_Value_1_0_EMPTY(TAGGED_UNION_PTR) ((TAGGED_UNION_PTR) && (TAGGED_UNION_PTR)->_tag_ == 0)
#define IS_UAVCAN__REGISTER_Value_1_0__STRING(TAGGED_UNION_PTR) ((TAGGED_UNION_PTR) && (TAGGED_UNION_PTR)->_tag_ == 1)
#define IS_UAVCAN__REGISTER_Value_1_0_UNSTRUCTURED(TAGGED_UNION_PTR) ((TAGGED_UNION_PTR) && (TAGGED_UNION_PTR)->_tag_ == 2)
#define IS_UAVCAN__REGISTER_Value_1_0_BIT(TAGGED_UNION_PTR) ((TAGGED_UNION_PTR) && (TAGGED_UNION_PTR)->_tag_ == 3)
#define IS_UAVCAN__REGISTER_Value_1_0_INTEGER64(TAGGED_UNION_PTR) ((TAGGED_UNION_PTR) && (TAGGED_UNION_PTR)->_tag_ == 4)
#define IS_UAVCAN__REGISTER_Value_1_0_INTEGER32(TAGGED_UNION_PTR) ((TAGGED_UNION_PTR) && (TAGGED_UNION_PTR)->_tag_ == 5)
#define IS_UAVCAN__REGISTER_Value_1_0_INTEGER16(TAGGED_UNION_PTR) ((TAGGED_UNION_PTR) && (TAGGED_UNION_PTR)->_tag_ == 6)
#define IS_UAVCAN__REGISTER_Value_1_0_INTEGER8(TAGGED_UNION_PTR) ((TAGGED_UNION_PTR) && (TAGGED_UNION_PTR)->_tag_ == 7)
#define IS_UAVCAN__REGISTER_Value_1_0_NATURAL64(TAGGED_UNION_PTR) ((TAGGED_UNION_PTR) && (TAGGED_UNION_PTR)->_tag_ == 8)
#define IS_UAVCAN__REGISTER_Value_1_0_NATURAL32(TAGGED_UNION_PTR) ((TAGGED_UNION_PTR) && (TAGGED_UNION_PTR)->_tag_ == 9)
#define IS_UAVCAN__REGISTER_Value_1_0_NATURAL16(TAGGED_UNION_PTR) ((TAGGED_UNION_PTR) && (TAGGED_UNION_PTR)->_tag_ == 10)
#define IS_UAVCAN__REGISTER_Value_1_0_NATURAL8(TAGGED_UNION_PTR) ((TAGGED_UNION_PTR) && (TAGGED_UNION_PTR)->_tag_ == 11)
#define IS_UAVCAN__REGISTER_Value_1_0_REAL64(TAGGED_UNION_PTR) ((TAGGED_UNION_PTR) && (TAGGED_UNION_PTR)->_tag_ == 12)
#define IS_UAVCAN__REGISTER_Value_1_0_REAL32(TAGGED_UNION_PTR) ((TAGGED_UNION_PTR) && (TAGGED_UNION_PTR)->_tag_ == 13)
#define IS_UAVCAN__REGISTER_Value_1_0_REAL16(TAGGED_UNION_PTR) ((TAGGED_UNION_PTR) && (TAGGED_UNION_PTR)->_tag_ == 14)

#ifdef __cplusplus
} /* End extern "C" */
#endif /* __cplusplus */
#endif /* UAVCAN__REGISTER_VALUE */
```


src/nunavut/lang/c/templates/_fields.j2
```diff
@@ -0,0 +1,11 @@
+{%- for field in composite_type.fields %}
+{%- if field is not padding %}
+    {{ field.data_type | declaration }} {{ field | id }}
+    {%- if field.data_type is ArrayType -%}
+        [{{ field.data_type.capacity }}]
+    {%- endif -%};
+{%- if field is VariableLengthArrayType %}
+    {{ typename_unsigned_length }} {{ field | id }}_length;
```
pavel-kirienko, line 8: I spent a considerable amount of time trying to find the source that advised to put length *before* the array in structures and parameter list declarations but I couldn't find it. I recall reading that some time ago possibly in the context of C11. Can we please put the length before the array? I might find the source later.

src/nunavut/lang/c/templates/_fields_as_union.j2
```diff
@@ -0,0 +1,9 @@
+    
+    {{ composite_type.tag_field_type | declaration }} _tag_;
+    union {
```
pavel-kirienko, line 3:
```suggestion
    union
    {
```

src/nunavut/lang/c/templates/_fields.j2
```diff
@@ -0,0 +1,11 @@
+{%- for field in composite_type.fields %}
+{%- if field is not padding %}
```
pavel-kirienko, line 2: There is `fields_except_padding` just for this. Applies to the union as well.

verification/c/suite/test_compiles.cpp
```diff
@@ -0,0 +1,57 @@
...
+#include "uavcan/time/TimeSystem_0_1.h"
+#include "uavcan/primitive/Empty_1_0.h"
+#include "uavcan/_register/Value_1_0.h"
+#include "uavcan/primitive/String_1_0.h"
+#include "uavcan/_register/Access_1_0.h"
+
+/**
+ * Verify that a simple struct type compiles.
+ */
+TEST(GeneralTests, StructCompiles) {
```
pavel-kirienko, line 16:
```suggestion
TEST(GeneralTests, StructCompiles)
{
```

I think we could use clang-format here

src/nunavut/lang/c/templates/UnionType.j2
```diff
@@ -0,0 +1,10 @@
+{% extends "Header.j2" %}
+{%- block object -%}
+{% set composite_type = T -%}
+{% include '_composite_type.j2' %}
+{%- for field in composite_type.fields -%}
+{%- if field is not padding %}
+#define IS_{{ composite_type | full_reference_name }}_{{ field.name | macrofy(partial_token=True) }}(TAGGED_UNION_PTR) ((TAGGED_UNION_PTR) && (TAGGED_UNION_PTR)->_tag_ == {{ loop.index0 }})
```
pavel-kirienko, line 7: See my comment about the entity naming pattern below. I suggest using the DSDL name as the prefix everywhere for consistency so instead of `IS_UAVCAN__REGISTER_Value_1_0_EMPTY(...)` we get `uavcan_register_Value_1_0_is_empty(...)`.

src/nunavut/lang/c/templates/_composite_type.j2
```diff
@@ -0,0 +1,67 @@
...
+{%- endfor %}
+{% import '_init_macros.j2' as init_macros %}
+/**
+ * Initialize an {{ composite_type | declaration }} instance to default values.
+ * No memory is allocated within this method.
+ *
+ * @param inout_instance A structure instance to Initialize.
+ * @return The inout_instance pointer for fluency.
+ */
+inline {{ composite_type | declaration }}* init_{{ composite_type | declaration | to_snake_case }}({{ composite_type | declaration }}* inout_instance)
```
pavel-kirienko, line 46:
```suggestion
inline {{ composite_type | declaration }}* init_{{ composite_type | declaration | to_snake_case }}({{ composite_type | declaration }}* const out_instance)
```

Shouldn't it be `out_instance` considering that the value is overridden unconditionally?

src/nunavut/lang/c/__init__.py
```diff
@@ -207,11 +209,84 @@ def filter_macrofy(value: str) -> str:
...
+        # which allows invalid identifiers to be generated.
+
+        # Given
+        template = '#ifndef FOO{{ "_invalid.register" | macrofy(partial_token=True) }}'
+
+        # then
+        rendered = '#ifndef FOO_INVALID__REGISTER'
+
+        # Stropping is still enabled so tokens are still stropped but the partial_token
+        # parameter skips stropping the final token.
```
pavel-kirienko, line 266: Double underscores cannot appear in user-defined identifiers. Maybe we should just replace them with a single underscore?

src/nunavut/lang/c/templates/UnionType.j2
pavel-kirienko, line 7: There is also a question of type safety and MISRA compliance. High-integrity guidelines generally prohibit the usage of function-like macros. Regardless of that, the macro-based design is prone to undetected type errors if the user feeds the pointer to a wrong union to the macro resulting in hard-to-detect runtime failures. Should we not use inline functions instead?

pavel-kirienko:
Also, things that I forgot to mention in the review:

1. High-integrity guidelines generally require that all return values are to be either used or explicitly cast to void. Considering that, the current design of the initialization function is suboptimal because it will require the user to cast its return value to void explicitly which is inconvenient. I recommend changing the type of the function to make it return nothing.

2. Standard library headers should be included last.

src/nunavut/lang/c/templates/_composite_type.j2
```diff
@@ -0,0 +1,67 @@
+{%- for constant in composite_type.constants %}
+#define {{ composite_type | c.macrofy }}_{{ constant.name | c.macrofy }} {{ constant | constant_value }}
+{%- endfor %}
+{%- if composite_type.fields | length == 0 %}
+{#- 
+    C++ doesn't tolerate empty types. To maintain compatibility we define any empty
```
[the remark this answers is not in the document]
thirtytwobits: I'll have to dig it up but I got C++ compatibility warnings when I tried to compile empty structs.

src/nunavut/lang/c/templates/_composite_type.j2
```diff
@@ -0,0 +1,67 @@
...
+{%- endif %}
+{%- for field in composite_type.fields %}
+{%- if field.data_type is ArrayType %}
+
+/* ArrayType macros for {{ field.name }} field of {{ composite_type | full_reference_name }}. */
+#define {{ composite_type | full_reference_name }}_{{ field.name | macrofy(partial_token=True) }}_ARRAY_CAPACITY ({{ field.data_type.capacity }}U)
+#define {{ composite_type | full_reference_name }}_{{ field.name | macrofy(partial_token=True) }}_ARRAY_IS_VARIABLE_LENGTH {% if field.data_type is VariableLengthArrayType -%}
+    (1U)
+{%- else -%}
+    (0U)
```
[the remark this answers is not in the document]
thirtytwobits: I make the use of stdbool optional. I can switch on that there and generate true/false if the default configuration to use stdbool is turned off.

src/nunavut/lang/c/templates/_composite_type.j2
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * Initialize an {{ composite_type | declaration }} instance to default values.
+ * No memory is allocated within this method.
+ *
+ * @param inout_instance A structure instance to Initialize.
+ * @return The inout_instance pointer for fluency.
+ */
+inline {{ composite_type | declaration }}* init_{{ composite_type | declaration | to_snake_case }}({{ composite_type | declaration }}* inout_instance)
+{
+    if (inout_instance)
```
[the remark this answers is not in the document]
thirtytwobits: I'll have to add `<stdlib.h>` to the include generator if I'm going to use NULL

src/nunavut/lang/c/__init__.py
```diff
@@ -207,11 +209,84 @@ def filter_macrofy(value: str) -> str:
...
 
     :returns: A valid C preprocessor identifier token.
     """
-    return value.replace(' ', '_').replace('.', '_').upper()
+    tokens = str(value).replace(' ', '.').split('.')
+    if language.enable_stropping:
+        stropped_tokens = list(map(functools.partial(filter_id, language), tokens))
+    else:
+        stropped_tokens = tokens
+    uppercase_stropped_tokens = [c.upper() for c in stropped_tokens]
```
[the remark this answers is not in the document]
thirtytwobits: I was using the UPPERCASE_NAMESPACE_Type convention as we discussed on the dev call two weeks ago. I'm not married to it. I can switch to utilizing the types without modifying case.

src/nunavut/lang/c/templates/_composite_type.j2
thirtytwobits: I always consider the "in" part meaning you have to provide valid memory in. Is there a better governing school of thought you could reference here?

src/nunavut/lang/c/templates/_composite_type.j2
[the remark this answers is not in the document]
thirtytwobits: Found it:
```
error: empty struct has size 0 in C, size 1 in C++ [-Werror=c++-compat]
```

i.e. -> https://godbolt.org/z/6bLUgC

src/nunavut/lang/c/templates/UnionType.j2
thirtytwobits: inline functions it is then.

thirtytwobits:
> Also, things that I forgot to mention in the review:
> 
> 1. High-integrity guidelines generally require that all return values are to be either used or explicitly cast to void. Considering that, the current design of the initialization function is suboptimal because it will require the user to cast its return value to void explicitly which is inconvenient. I recommend changing the type of the function to make it return nothing.

OKAAAAY. My dreams of writing object-c-ish statements like:

```
Foo* a = init_foo(allocate_foo());
```

have been crushed!

src/nunavut/lang/c/templates/_composite_type.j2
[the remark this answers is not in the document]
pavel-kirienko: Ah indeed. I think we should keep things consistent and ensure that every generated type is a struct, otherwise, in C++ templated code we might run into complicated corner cases where one might want to use type traits.

Maybe like this?

    typedef struct {{ composite_type | full_reference_name }} { uint8_t _dummy_; } {{ composite_type | full_reference_name }};

src/nunavut/lang/c/__init__.py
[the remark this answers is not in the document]
pavel-kirienko: I might have missed that, sorry. Let's use the original case.

src/nunavut/lang/c/templates/_composite_type.j2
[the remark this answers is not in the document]
pavel-kirienko: Indeed, but MISRA et al require that the implicit conversion of a pointer to boolean is not permitted. Are there any known platforms that do not provide stdlib.h?

src/nunavut/lang/c/templates/_composite_type.j2
pavel-kirienko: That would imply that there cannot be a pure `out` parameter because you always have to provide valid storage for the output. Related: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/out-parameter-modifier

src/nunavut/lang/c/templates/_composite_type.j2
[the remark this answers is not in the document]
thirtytwobits: Not that I'm aware of. It's just that I need to generate the include. No problem. I'll make this change.

src/nunavut/lang/c/templates/_composite_type.j2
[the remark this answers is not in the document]
thirtytwobits: sure.

src/nunavut/lang/c/templates/_composite_type.j2
thirtytwobits: Yep. Thanks. I'll change this to just "out".

src/nunavut/lang/c/__init__.py
```diff
@@ -206,12 +208,65 @@ def filter_macrofy(value: str) -> str:
...
+        config_overrides = {'nunavut.lang.c': {'enable_stropping': False }}
+        lctx = configurable_language_context_factory(config_overrides, 'c')
+        jinja_filter_tester(filter_macrofy, template, rendered, lctx)
 
     :param str value: The value to transform.
 
     :returns: A valid C preprocessor identifier token.
     """
-    return value.replace(' ', '_').replace('.', '_').upper()
+    macrofied_value = filter_to_screaming_snake_case(str(value))
```
pavel-kirienko, line 265: Should we not apply consistent naming throughout? Instead of

```
#define UAVCAN_TIME_TIME_SYSTEM_0_1_MONOTONIC_SINCE_BOOT 0
```

have

```
#define uavcan_time_TimeSystem_0_1_MONOTONIC_SINCE_BOOT 0
```

I see that some other entities have been renamed already as suggested during the last review but the constants stand out still.

src/nunavut/lang/c/templates/UnionType.j2
```diff
@@ -0,0 +1,13 @@
+{% extends "Header.j2" %}
+{%- block object -%}
+{% set composite_type = T -%}
+{% include '_composite_type.j2' %}
+{%- for field in composite_type.fields -%}
+{%- if field is not padding %}
+inline {{ typename_boolean }} is_{{ composite_type | full_reference_name }}_{{ field.name }}(const {{ composite_type | full_reference_name }}* const tagged_union_ptr)
```
pavel-kirienko, line 7:
```suggestion
inline {{ typename_boolean }} {{ composite_type | full_reference_name }}_is_{{ field.name }}(const {{ composite_type | full_reference_name }}* const tagged_union_ptr)
```

Consistent token ordering.

src/nunavut/lang/c/__init__.py
thirtytwobits: The constants are macros as such there is a consistency that all macros are SCREAMING_SNAKE_CASE to communicate to the developer an important difference in their nature. I believe this is the only deviation from the modified snake case used elsewhere.

src/nunavut/jinja/TimeSystem_0_1.h
```diff
@@ -0,0 +1,67 @@
+/*
```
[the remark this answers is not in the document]
thirtytwobits: yes. Sorry. user error. will remove.

src/nunavut/lang/c/__init__.py
pavel-kirienko: Whatever we choose it will end up looking odd -- either macros that don't look like macros or macros that don't look like the things they are supposed to represent. I am inclined to break the coding convention rather than DSDL convention here because I think it adds clarity and removes the risk (which is admittedly marginal though) of case-sensitive name conflicts.

Also, consider the issues with metric suffixes. We discussed it a while ago on the forum and the convention made its way into the specification: if an attribute has a metric suffix, the case of the suffix is independent of the case of the attribute itself. For example, there may be `MAX_ENERGY_mJ` where `mJ` stands for millijoule. Do you see how the screaming snake case transformation might introduce an error of six orders of magnitude? Scary :snake:

src/nunavut/lang/c/templates/_fields.j2
thirtytwobits: pish-posh. Okay

src/nunavut/lang/c/templates/_fields.j2
thirtytwobits: Oh nice!

PR #125, opened 2020-08-10
Adding support header templates 

thirtytwobits:
WIth this change the support headers can be `.j2` Jinja templates. When they are rendered there will be no `T` available (i.e. these are not generated per-type) but the target language environment will be available including language options like `option_target_endianness`.

### DIFF
--- a/src/nunavut/__init__.py
+++ b/src/nunavut/__init__.py
@@ -41,7 +41,7 @@ Putting this all together, the typical use of this library looks something like 
     from pydsdl import read_namespace
     from nunavut import build_namespace_tree
     from nunavut.lang import LanguageContext
-    from nunavut.jinja import Generator
+    from nunavut.jinja import DSDLCodeGenerator
 
     # parse the dsdl
     compound_types = read_namespace(root_namespace, include_paths)
@@ -56,7 +56,7 @@ Putting this all together, the typical use of this library looks something like 
                                           language_context)
 
     # give the root namespace to the generator and...
-    generator = Generator(root_namespace)
+    generator = DSDLCodeGeneratorroot_namespace)
 
     # generate all the code!
     generator.generate_all()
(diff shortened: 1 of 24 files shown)


thirtytwobits:
I think this works but I need to write more tests to be sure. The happy case appears functional so I'm posting this PR to let you and @davidlenfesty know of the design.

src/nunavut/__init__.py
```diff
@@ -56,7 +56,7 @@
                                           language_context)
 
     # give the root namespace to the generator and...
-    generator = Generator(root_namespace)
+    generator = DSDLCodeGeneratorroot_namespace)
```
pavel-kirienko, line 59: wtf

src/nunavut/lang/cpp/support/serialization.j2
```diff
@@ -14,7 +14,7 @@ namespace nunavut
 {
 namespace support
 {
-
+// Target Endianess: {{ option_target_endianness }}
```
pavel-kirienko, line 17: Where does this option come from exactly? I am asking because I intended to verify that the available options are "little" and "any" (rather than "little" and "big").

src/nunavut/__init__.py
thirtytwobits: Screen Shot 2020-08-12 at 10 27 50 AM

src/nunavut/lang/cpp/support/serialization.j2
thirtytwobits: Added 'any' as a valid option. It's up to the template author to consume these values, ultimately. If I'm wrong I'll eat my hat but I will not, under any circumstances, eat my dog.

pavel-kirienko: Another useful phrase to have in your Russian lexicon is "[let there be] a dog with (subject)" ("пёс с ним"), which conveys indifference towards the subject.

Issue #126, opened 2020-08-14
Option to allow allow_unregulated_fixed_port_id (0.3.4)


nacansino:
**Nunavut Version**
0.3.4

**Details:**
As the title says. 

I encountered the error `raise UnregulatedFixedPortIDError` when generating C-headers against a DSDL namespace root:
\root
|---100.PointXY.1.0.uavcan
|---MyService.1.0.uavcan

100 should be counted as unregulated identifiers as per Table 5.1 in the UAVCAN specification.

I don't know if this was intended or not, but it seems that Nunavut is calling `pydsdl` with the default argument `allow_unregulated_fixed_port_id: bool=False`

inside cli.py:
```python
...
    #
    # nunavut : parse
    #
    type_map = pydsdl.read_namespace(args.root_namespace, extra_includes)
...
```

**Possible solution:**
Expose the allow_unregulated_fixed_port_id argument on nnvg?

thirtytwobits:
Oh weird. I thought I did that already. This is an easy thing to add.

thirtytwobits:
released

PR #127, opened 2020-08-15
issue #126, adding allow unregulated arg to nnvg

thirtytwobits:
Adds feature requested by #126

### DIFF
--- a/src/nunavut/cli.py
+++ b/src/nunavut/cli.py
@@ -367,7 +367,7 @@ def _make_parser() -> argparse.ArgumentParser:
 
     ''').lstrip())
 
-    parser.add_argument('-upid', '--allow-unregulated-fixed-port-id',
+    parser.add_argument('--allow-unregulated-fixed-port-id',
                         action='store_true',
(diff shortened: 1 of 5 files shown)


src/nunavut/cli.py
```diff
@@ -365,6 +367,16 @@ def extension_type(raw_arg: str) -> str:
 
     ''').lstrip())
 
+    parser.add_argument('-upid', '--allow-unregulated-fixed-port-id',
```
pavel-kirienko, line 370:
```suggestion
    parser.add_argument('--allow-unregulated-fixed-port-id',
```

dangerous things should be hard to reach

thirtytwobits: Accepted but I thought you'd appreciate the pun (i.e. the missing "st")

pavel-kirienko: This pun requires a user manual. Or maybe it's just me -- I grew up in a town named Stupino which might have made me insensitive to this specific kind of jokes, not sure.

PR #128, opened 2020-08-15
Docs CSS: fix display on wide screens

pavel-kirienko:
I am not particularly opinionated about this but I was just working on moving the PyDSDL docs to ReadTheDocs and I thought that I should introduce the layout fix into Nunavut as well because this doesn't look right:

The effect of this change can be seen in the PyUAVCAN docs.

I wonder if I should propose this fix upstream...

### DIFF
--- a/docs/static/css/custom.css
+++ b/docs/static/css/custom.css
@@ -2,3 +2,6 @@ p.first {
     white-space: normal;
     width: 30em;
 }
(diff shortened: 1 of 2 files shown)


thirtytwobits:
Love it but Nunavut still requires patch version bumps for any change on mainline.

PR #131, opened 2020-08-22
[C serialization] Fix recursion issue in init_field

davidlenfesty:
This is my first kick at the can.

I haven't done any more than the simplest amount of testing.

The most obvious solution in my mind was to modify it to be more in line with how I did it in the serialization code, because you cannot always know the field "name", so it's easier to pass strings down than it is to extract semantics like that.

It *seems* to work on the BatteryPackParams type that it originally crashed on, but I haven't checked anything else,
and I'm especially concerned about how it'll handle multi-dimensional arrays.

### DIFF
--- a/src/nunavut/lang/c/templates/_init_macros.j2
+++ b/src/nunavut/lang/c/templates/_init_macros.j2
@@ -1,21 +1,29 @@
-{%- macro init_field(instance_ptr_token, field, index='') -%}
-    {%- if field is not padding -%}
-    {%- if field.data_type is CompositeType -%}
-        {{ field.data_type | declaration }}_init(&({{ instance_ptr_token }}->{{ field | id }}{{ index }}));
-    {%- elif field.data_type is FixedLengthArrayType -%}
-        {%- if field.data_type.element_type is PrimitiveType -%}
-        memset(&({{ instance_ptr_token }}->{{ field | id }}[0]), 0, sizeof({{ field.data_type | declaration }}) * {{ field.data_type.capacity }});
+{%- macro init_field(field, data_type=None, ref='out_instance->') -%}
+    {%- if field is None -%}
+        {%- set _data_type = data_type -%}
+        {%- set _out = ref -%}
+    {%- else -%}
+        {%- set _data_type = field.data_type -%}
+        {%- set _out = ref + (field | id) -%}
+    {%- endif -%}
+
+    {%- if field is not padding or field is None -%}
+    {%- if _data_type is CompositeType -%}
+        {{ _data_type | declaration }}_init(&({{ _out }}));
+    {%- elif _data_type is FixedLengthArrayType -%}
+        {%- if _data_type.element_type is PrimitiveType -%}
+        memset(&({{ _out }}[0]), 0, sizeof({{ _data_type | declaration }}) * {{ field.data_type.capacity }});
         {%- else -%}
-             {%- for index in range(field.data_type.capacity) %}
-                {{ init_field(instance_ptr_token, field, '[{}]'.format(index)) }}
+             {%- for index in range(_data_type.capacity) -%}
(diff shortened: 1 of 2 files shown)


src/nunavut/lang/c/templates/_init_macros.j2
```diff
@@ -1,21 +1,29 @@
-{%- macro init_field(instance_ptr_token, field, index='') -%}
-    {%- if field is not padding -%}
-    {%- if field.data_type is CompositeType -%}
-        {{ field.data_type | declaration }}_init(&({{ instance_ptr_token }}->{{ field | id }}{{ index }}));
-    {%- elif field.data_type is FixedLengthArrayType -%}
-        {%- if field.data_type.element_type is PrimitiveType -%}
-        memset(&({{ instance_ptr_token }}->{{ field | id }}[0]), 0, sizeof({{ field.data_type | declaration }}) * {{ field.data_type.capacity }});
+{%- macro init_field(field, data_type=None, ref='out_instance->') -%}
+    {%- if field is None -%}
```
thirtytwobits, line 2: Huh. I thought this didn't work in Jinja : https://jinja.palletsprojects.com/en/2.11.x/templates/#assignments ? Are you sure this is working like you expect it to? Do we need to use a namespace object?

src/nunavut/lang/c/templates/_init_macros.j2
```diff
@@ -1,21 +1,29 @@
...
+        {%- set _data_type = data_type -%}
+        {%- set _out = ref -%}
+    {%- else -%}
+        {%- set _data_type = field.data_type -%}
+        {%- set _out = ref + (field | id) -%}
+    {%- endif -%}
+
+    {%- if field is not padding or field is None -%}
+    {%- if _data_type is CompositeType -%}
+        {{ _data_type | declaration }}_init(&({{ _out }}));
```
thirtytwobits, line 12: the ref default is `out_instance->` which means if you have `init_field(None, _data_type.element_type)` this would generate `&(out_instance->)` which is invalid syntax.

src/nunavut/lang/c/templates/_init_macros.j2
```diff
@@ -1,21 +1,29 @@
...
+    {%- elif _data_type is FixedLengthArrayType -%}
+        {%- if _data_type.element_type is PrimitiveType -%}
+        memset(&({{ _out }}[0]), 0, sizeof({{ _data_type | declaration }}) * {{ field.data_type.capacity }});
         {%- else -%}
-             {%- for index in range(field.data_type.capacity) %}
-                {{ init_field(instance_ptr_token, field, '[{}]'.format(index)) }}
```
thirtytwobits, line 10: Ugh. I can't believe I wrote this line of code. Sorry. It's so obviously wrong now that I look at it.

thirtytwobits:
Okay. I whipped up tests that reproduce the problem here -> https://github.com/thirtytwobits/nunavut/tree/issue/130
If you rebase on master then pull in that branch (as a patch) then you should be able to verify that your fix works.

thirtytwobits:
@davidlenfesty , do you want me to take over this PR so you can concentrate on the serialization changes? It's really close now that we have the repro tests in-place.

davidlenfesty:
@thirtytwobits I definitely wouldn't mind that.

Note that this actually doesn't work as it stands either :P. Currently it will generate something like:

```
compound_type_init(in_instance->[0], blah blah);
```

Where it's supposed to generate:
```
compound_type_init(in_instance->field[0], blah blah);
```

This probably comes out in BatteryPackParams, and will likely be fixed if you refactor into two macros. I haven't had any time to work on it so that's all I've got there.

thirtytwobits:
Cool. Thanks for starting this. It was a big help to have the problem all isolated. I got it from here. I have a working prototype now and am just trying some more deeply recursive examples.

thirtytwobits:
Closing in favor of #135

Issue #152, opened 2020-10-31
add required extent or sealed keywords


thirtytwobits:
fix the tests now that pydsdl requires @extent or @sealed

Issue #153, opened 2020-11-01
pydsdl removal of parent_service attribute breaks Nunavut


thirtytwobits:
this was used by the `full_reference_name` filter.

thirtytwobits:
I was handling service type identifiers wrong and treating them as types instead of namespaces. I'm still not sure if service types should include their version numbers?

PR #155, opened 2020-11-02
Fixes for pydsdl changes

thirtytwobits:
We've changed both the requirement to have either @sealed or @extent and how we treat service types.

This fixes issue #152  and #153

### DIFF
--- a/src/nunavut/lang/c/__init__.py
+++ b/src/nunavut/lang/c/__init__.py
@@ -736,13 +736,14 @@ def filter_full_reference_name(language: Language, t: pydsdl.CompositeType) -> s
         import pydsdl
 
         my_obj = MagicMock()
-        my_obj.parent_service = None
+        my_obj.has_parent_service = False
         my_obj.version = MagicMock()
 
     .. code-block:: python
 
         # Given a type with illegal characters for C++
         my_obj.full_name = 'any.int.2Foo'
+        my_obj.full_namespace = 'any.int'
         my_obj.version.major = 1
         my_obj.version.minor = 2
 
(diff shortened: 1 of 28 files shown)


pavel-kirienko (approved):
It's interesting that you treat service types as namespaces here. Maybe it makes sense. In the future, we may change the spec to make request/response types reachable (make them non-Voldemort), in which case having the version number attached directly to the type rather than its parent would be sensible. PyDSDL is indifferent to this because it guarantees that request/response share metadata like version, deprecation status, source file path, etc. with their parent service type.

thirtytwobits:
> It's interesting that you treat service types as namespaces here. Maybe it makes sense. In the future, we may change the spec to make request/response types reachable (make them non-Voldemort), in which case having the version number attached directly to the type rather than its parent would be sensible. PyDSDL is indifferent to this because it guarantees that request/response share metadata like version, deprecation status, source file path, etc. with their parent service type.

Hmm. now I wonder if I'm doing the right thing here? I suppose I can restore the old behavior using the assumption that the request and response type's version is always the same as the parent service type and that the parent's name is the namespace part that contains the request and response types. Do you think that `foo_BarService_1_0_Request` is actually more correct then `foo_BarService_Response_1_0`? The nice thing about the way I was doing it before was the short references were `BarService_1_0_Request` instead of `Request_1_0`. The former created more descriptive names then the latter, obviously.

pavel-kirienko:
As per the current Spec, if you look hard enough, the correct approach is to say `BarService_1_0`, then `Request`, but the Spec doesn't *require* this, it never even touches this question because it is not relevant for wire interoperability. If we look into the future, then it is more likely (as I see things) that we will be converging towards `Request_1_0`.

> The nice thing about the way I was doing it before was the short references were BarService_1_0_Request instead of Request_1_0. The former created more descriptive names then the latter, obviously.

Meh, I don't think it matters at all. PyUAVCAN says just `Request`, not even mentioning the version numbers. In the end, there will always be a risk of collision if you use the short name only.

thirtytwobits:
Hmm. I think my error is actually that I'm generating named types for Request and Response instead of making them anonymous struts in a concrete service struct. The logic in this CR is correct then but it shouldn't be possible to actually generate types for Request and Response. Do you think I should open an issue against that?

Issue #159, opened 2020-11-03
Specify line post-processor per language


pavel-kirienko:
I need Nunavut to apply `LimitEmptyLines(1)` and `TrimTrailingWhitespace()` to generated C outputs, but at the moment it is only possible to do so via CLI. There should be a way to specify this via language defaults.

@thirtytwobits would you be available to implement this?

thirtytwobits:
Ah ya big baby. Okay. I'll add a way to automatically run LimitEmptyLines and TrimTrailingWhitespace but enabling these will slow down generation since we'll have to fallback to line buffered code generation. Can you not just write the jinja correctly?

pavel-kirienko:
> Can you not just write the jinja correctly?

I can but I don't understand why should I bother? Managing whitespaces in any non-trivial template is tedious work that requires attention and adjustments; it is easy to automate away completely so why not do so?

This task is now a low-priority one because I implemented a workaround in https://github.com/UAVCAN/nunavut/commit/316b4f196727126a3d6c4d85105eadf6ec2a1d5a. Right now https://github.com/UAVCAN/nunavut/issues/157 is more critical.

pavel-kirienko:
*(I also prefer garbage collected languages and cars with automatic transmission)*

thirtytwobits:
> _(I also prefer garbage collected languages and cars with automatic transmission)_

You might find it funny to note that I despise garbage collected languages and I learned to drive using a standard transmission (and...I...liked it!)

aentinger:
Everyone likes standard transmission, the fast and furious franchise saw to that :wink: That being said, I don't really dislike garbage collected languages but I don't trust them :wink: (Sorry, but I couldn't resist commenting when @thirtytwobits above comment flowed into my inbox.)

PR #161, opened 2020-11-10
C serialization codegen

### LINKED ISSUE #112: C serialization support
Depends on issue #111 : generate serialization routines and support header to C POD types unless the `-pod` switch is set.

### LINKED ISSUE #114: Message size definition bits or bytes?
I've been verifying the PX4 UAVCAN v1 implementation and noticed a bug in that implementation that the messages were to big (8 times to be exactly). It seems this was caused by confusion between the message size generated using nunavut with following template.

`static constexpr size_t SIZE = {{total_size.value}};`

And the payload size definition used in libcanard; to clarify the message size was bits and the payload is defined in bytes.

Question: what would be the best solution to solve this issue?

1. Change total_size.value definition in nunavut
2. Change the template by dividing by 8 e.g.
`static constexpr size_t SIZE = {{total_size.value // 8}};`
3. Divide the size by 8 when assigning the payload size e.g.
`.payload_size   = uavcan::node::Heartbeat_1_0::SIZE / 8,`

My personal preference would be to go for option 1 or option 2, mostly because I would expect that a size variable would be in bytes; but that's just a personal preference.

### LINKED ISSUE #143: C primitive serialization templates do not respect the cast mode
https://github.com/davidlenfesty/nunavut/blob/d1ba583c131490b2a4f8ca594b14ca7dee4f762c/src/nunavut/lang/c/templates/_serialize.j2#L170-L182

When serializing a primitive whose cast mode is `saturated` (which is the default), the serializer should check the range first and clip the value as necessary. This applies to `intx`, `uintx`, `float16`.

### LINKED ISSUE #145: Support delimited serialization
The details are described in Specification v1.0-beta.

Until delimited serialization is supported, generated code should trigger an assertion failure when one attempts to (de)serialize a nested delimited composite.

### LINKED ISSUE #160: Ready for Beta Testing
Is it ready for beta testing?  Where can we sign up for beta testing?

pavel-kirienko:
I have picked up where @davidlenfesty and @thirtytwobits have left off and made the logic compliant with the beta spec and reasonably well-tested. This is intended to be usable already.

The following items are missing:

- [ ] https://github.com/UAVCAN/nunavut/issues/157 (need help from @thirtytwobits)
- [ ] https://github.com/UAVCAN/nunavut/issues/159 (need help from @thirtytwobits)
- [ ] https://github.com/UAVCAN/nunavut/issues/162 (need help from @thirtytwobits)
- [x] Buildkite keeps aborting the build after 15 minutes despite the updated config; @thirtytwobits can you please give it a look?

Fixes #160
Fixes #145
Fixes #143
Fixes #114
Fixes #112

@hugo-a-garcia @aentinger @PetervdPerk-NXP  volunteer for testing today!

image

Example of generated code: [c.tar.gz](https://github.com/UAVCAN/nunavut/files/5519900/c.tar.gz)


pavel-kirienko:
Here's the usage example for testing:

```bash
git clone https://github.com/UAVCAN/nunavut --branch c-serialization-corrections
cd nunavut
pip install .
cd -
git clone https://github.com/UAVCAN/public_regulated_data_types/
nnvg --target-language c --pp-max-emptylines=1 --pp-trim-trailing-whitespace --target-endianness=little --enable-serialization-asserts public_regulated_data_types/uavcan
```

Options `--pp-max-emptylines=1 --pp-trim-trailing-whitespace` will become unnecessary after #159 is fixed.

Option `--target-endianness=little` may be omitted, which will result in endianness-agnostic code that is portable but slower and possibly larger.

Option `--enable-serialization-asserts` injects cross-checks in the generated serialization routines which slows them down and increases the ROM footprint but protects against possible memory corruption and detects some of the possible issues in the generated code. It is recommended to always enable it.

thirtytwobits:
FYI: I've got three commits I've rebased on your "corrections" branch here -> https://github.com/thirtytwobits/nunavut/commit/7ac4372d97e35344efbdd3c24406ff0c8de678ad

Still working on finishing this out. One thing I forgot to ask, do we need to use QEMU to test big-endianess? I don't see the point in providing this switch if we can't actually test the code.

pavel-kirienko:
> do we need to use QEMU to test big-endianess? I don't see the point in providing this switch if we can't actually test the code.

That would be nice. The switch makes sense even without the test; we could equip such endianness-agnostic code with some warning that would trip when executed saying that the feature is untested (like it was done in libuavcan v0).

How difficult would it be to (eventually) expand the QEMU-based suite to support AVR or some other commonly used 8-bit architecture?

thirtytwobits:
> > do we need to use QEMU to test big-endianess? I don't see the point in providing this switch if we can't actually test the code.
> 
> That would be nice. The switch makes sense even without the test; we could equip such endianness-agnostic code with some warning that would trip when executed saying that the feature is untested (like it was done in libuavcan v0).
> 
> How difficult would it be to (eventually) expand the QEMU-based suite to support AVR or some other commonly used 8-bit architecture?

Trivial. QEMU already supports this.

TSC21:
@pavel-kirienko maybe you want to rebase @thirtytwobits's changes?

thirtytwobits:
> @pavel-kirienko maybe you want to rebase @thirtytwobits's changes?

Ack, hangon. Just realized there's a regression. Fixing...

thirtytwobits:
All done! @pavel-kirienko , you can rebase now and you get my new verify.py script to help you along. Sorry for that trouble.

.buildkite/verify_c.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
thirtytwobits, line 1: this will run everything in serial. If you define this matrix in the pipline.yml then we can run builds in parallel.

thirtytwobits: Fixed this myself.

thirtytwobits:
Let's merge this one and take all this change then work on the additional items listed in this PR.

thirtytwobits (approved):
I think this is a solid iteration. Let's take it and continue to work from here.

PR #168, opened 2020-11-29
Use POSIX-style paths in #includes

### LINKED ISSUE #129: Generated path on #include uses backslashes when generated on Windows
**Nunavut Version:**
0.3.4

**Description of the problem**
As the title says.
I generated C headers on a Windows machine. All was well except today when I moved into my workstation that uses Linux and pulled the headers from an online repo where the build failed miserably due to unresolved include paths.

**Proposed Solution:**
I suggest we use forward slashes in our #include paths regardless of the host generator.

edwinbalani:
The `__str__()` method of a `pathlib.Path` [is platform-dependent](https://docs.python.org/3.8/library/pathlib.html#operators).  We were using this when generating `#include` directives, but to produce host-independent C/C++ code we probably shouldn't (e.g. so that code generated on Windows can compile on a Linux system without Windows-style).

The current behaviour ultimately stems from the fact that the implementation presented at `pathlib.Path` depends on the interpreter host platform.  These are individually available as the `PosixPath` and `WindowsPath` classes.

An alternative to this PR would be to use `pathlib.PosixPath` (or `PurePosixPath`) around the code explicitly, rather than relying on platform-dependent behaviour.  However, it would probably be a much bigger diff than this tiny change.  I think keeping things as platform-native `Path` objects would be better, for instance so that the same path variable can be used to query the real filesystem without surprises.

Closes #129.

### DIFF
--- a/src/nunavut/lang/_common.py
+++ b/src/nunavut/lang/_common.py
@@ -39,7 +39,7 @@ class IncludeGenerator(DependencyBuilder):
             namespace_path = pathlib.Path('')
             for namespace_part in self._language.support_namespace:
                 namespace_path = namespace_path / pathlib.Path(namespace_part)
-            path_list += [str(namespace_path / pathlib.Path(p.name).with_suffix(output_extension))
+            path_list += [(namespace_path / pathlib.Path(p.name).with_suffix(output_extension)).as_posix()
                           for p in self._language.support_files]
 
         prefer_system_includes = bool(self._language.get_config_value_as_bool('prefer_system_includes', False))
--- a/src/nunavut/lang/_common.py
+++ b/src/nunavut/lang/_common.py
@@ -98,7 +98,7 @@ class IncludeGenerator(DependencyBuilder):
     def _make_path(self, dt: pydsdl.CompositeType, output_extension: str) -> str:
         short_name = self._short_reference_name_filter(self._language, dt)
         ns_path = pathlib.Path(*self._make_ns_list(dt)) / pathlib.Path(short_name).with_suffix(output_extension)
-        return str(ns_path)
+        return ns_path.as_posix()
 
     def _make_ns_list(self, dt: pydsdl.SerializableType) -> typing.List[str]:
         if self._language.enable_stropping:
--- a/src/nunavut/version.py
+++ b/src/nunavut/version.py
@@ -3,6 +3,6 @@
 # This software is distributed under the terms of the MIT License.
 #
 
-__version__ = "0.6.2"
+__version__ = "0.6.3"
 
 __license__ = 'MIT'


pavel-kirienko (changes requested):
Thanks, @edwinbalani! It looks good but please bump the patch version number in `version.py` to let our CI/CD setup release the new version automatically after merge.

edwinbalani:
> please bump the patch version number in `version.py` to let our CI/CD setup release the new version automatically after merge.

Thanks - I'll do so.  I think CONTRIBUTING.md could be clearer on this, since it currently only says to do this when "committing to master".  In one interpretation, I'm contributing to my branch at the moment :)

Do you think there are any situations where a PR _shouldn't_ bump the version number, and are they rarer than the ones where a contributor should?  If so, then perhaps the guidance could be reworded to say you normally should bump as part of your PR changes, and outline situations where it shouldn't be done.

I ought to test this change informally on Windows to check that the behaviour's actually changed; should be able to do this today too.

pavel-kirienko:
> Do you think there are any situations where a PR shouldn't bump the version number, and are they rarer than the ones where a contributor should? If so, then perhaps the guidance could be reworded to say you normally should bump as part of your PR changes, and outline situations where it shouldn't be done.

I don't think there is a sensible scenario where you shouldn't but we should wait for @thirtytwobits to confirm.

> I ought to test this change informally on Windows to check that the behaviour's actually changed; should be able to do this today too.

Speaking of which, there's #60

edwinbalani:
It seems the changes haven't quite done the trick.  Where we used to get this:
```
#include <nunavut\support\serialization.h>
#include <uavcan\_register\Name_1_0.h>
#include <uavcan\_register\Value_1_0.h>
#include <uavcan\time\SynchronizedTimestamp_1_0.h>
// ...
```
we now get
```
#include <nunavut/support/serialization.h>
#include <uavcan\_register\Name_1_0.h>
#include <uavcan\_register\Value_1_0.h>
#include <uavcan\time\SynchronizedTimestamp_1_0.h>
// ...
```

This is in the generated header for uavcan.register.Access.1.0, using `nnvg -e .h .\public_regulated_data_types\uavcan`, with Python 3.9.0 and a fresh clone of the types.

So I think there are more potential instances of `str()` that need stamping out.

pavel-kirienko:
Adding a test that ensures that there are no `\`-s would help here. It is expected to be redundant on GNU/Linux but once we support Windows in the CI it will help us avoid regressions.

edwinbalani:
...just one bad `str()`, which has hanging around at the bottom of the same file 😁.  I've also bumped the version number, and I think this is good to go now.

I've since learnt more about nnvg, and changed how I invoke it.  With `nnvg -l c -I .\public_regulated_data_types\uavcan\ -O reg_gen .\public_regulated_data_types\reg` I get nice-looking output:
```c
// reg.drone.phy.acoustics.Note.1.0
#include <nunavut/support/serialization.h>
#include <uavcan/si/unit/duration/Scalar_1_0.h>
#include <uavcan/si/unit/frequency/Scalar_1_0.h>
#include <uavcan/si/unit/power/Scalar_1_0.h>
#include <stdlib.h>
```

edwinbalani:
> Adding a test that ensures that there are no `\`-s would help here.

This should be straightforward, right? (Famous last words...)  We probably just need to check that lines starting with '#include' don't have a '\'.

pavel-kirienko:
That's one way to handle it, yes. Maybe you can narrow the scope of the test such that it only involves the IncludeGenerator itself without having to generate the whole file, but ultimately it doesn't matter much, I think.

thirtytwobits (approved):
Thanks!

edwinbalani:
I think [`test.gentest_filter.test_filters:test_python_filter_includes`](https://github.com/UAVCAN/nunavut/blob/a4147c54db014e522dbf975eb390ed0dfe7e4254/test/gentest_filters/test_filters.py#L192) actually covers us here, by hard-asserting that the output of the IncludeGenerator exactly matches a given list, so no new test should be needed.  It's only testing the C++ case, but both it and the C language support use the same IncludeGenerator, unsurpisingly.

So whenever CI on Windows happens, this should hopefully be one existing test that doesn't start failing out of the blue :sweat_smile:

thirtytwobits:
My change is blocked behind this one. @pavel-kirienko , is there more you need here?

pavel-kirienko (approved):

pavel-kirienko:
Sorry, I didn't realize that you were waiting for my review.

PR #177, opened 2020-12-23
Differentiate between stable/experimental languages in API and CLI

### LINKED ISSUE #172: Put C++ and js behind an "experimental" flag

edwinbalani:
A second approach to denoting 'experimental' languages (i.e. anything that's not C or Python for v1.0).

Controversially, I've chosen for the CLI to default to supplying `include_experimental_languages` as `False` when creating a LanguageContext, although in `LanguageContext.__init__` itself the default is `True`.  This would mean that most of the test code (and any Pythonic users of nunavut rather than CLI users) will have experimental languages available by default -- let me know how you feel about this.  My main motivations for that choice are that the bulk of the test code, which uses C++ and JSON outputs, can be left intact this way, and that the original issue #172 seems concerned mainly with CLI users.

@thirtytwobits please correct me if wrong - is Python stable, or is it just C right now?

Resolves #172.
Closes #173 by obsolence.

### DIFF
--- a/src/nunavut/cli.py
+++ b/src/nunavut/cli.py
@@ -76,7 +76,8 @@ def _run(args: argparse.Namespace, extra_includes: typing.List[str]) -> int:  # 
         args.output_extension,
         args.namespace_output_stem,
         omit_serialization_support_for_target=args.omit_serialization_support,
-        language_options=language_options)
+        language_options=language_options,
+        include_experimental_languages=args.experimental_languages)
 
     #
     # nunavut: inferred target language from extension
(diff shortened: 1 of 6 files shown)


edwinbalani:
Only failing check now is a lint:

```
src/nunavut/lang/__init__.py:427:5: C901 'LanguageContext.__init__' is too complex (12)
    def __init__(self,
    ^
1     C901 'LanguageContext.__init__' is too complex (12)
```

Indicates too many branches... I'll see whether I can make things a bit less 🎄-ish

thirtytwobits (changes requested):
We just need to update version.py with a new, minor revision. Can you do that and close your other PR and I'll accept and merge. Thanks for doing this!

edwinbalani:
> We just need to update version.py with a new, minor revision. Can you do that and close your other PR and I'll accept and merge. Thanks for doing this!

Great - no worries, I'm glad I could get one of the v1.0 tasks done!

By minor revision, I'll assume you mean the 'z' part of 'x.y.z'.  Commit incoming...

edwinbalani:
On second thoughts, I trust that you meant the 'y', since this is a more significant change in program behaviour than what warrants a patch rev (_and_ especially since we're technically still in v0, it's probably best placed in a minor revision rather than a major).

thirtytwobits (approved):

PR #186, opened 2021-02-08
Fix missing .format in experimental language warning

bbworld1:
Really small PR here. Noticed that there was a missing .format() on this warning, so I added it.

Not sure how strict the issue -> PR etiquette is on this project - this is just a one-line bugfix, but if the procedure is to open an issue then I can open one.


thirtytwobits:
Oh wow. How did our build miss this? That tells me we have a gap in our testing. I'll accept this and look into that. Thanks.

thirtytwobits:
Added #187 to follow up with tests.

PR #194, opened 2021-03-14
Issue #91 progress.

thirtytwobits:
This is a checkpoint merge to put a firebreak in a growing change-set. Most all of these changes are behind the --experimental-language flag so there's little risk. There are a couple of fixes in the core python library that should not break anybody we know of.

Highlights:
1. Mostly complete generation of PoD C++
2. Significant progress on the variable-length-array problem. Included is a support library implementation that I'll hook up when creating the serialization routines.
3. Testing using Python 3.9 is added.
4. A new comment section in the C headers will document the python version and environment used to generated the types.


src/nunavut/lang/cpp/templates/_fields_as_union.j2
```diff
@@ -1,17 +1,22 @@
...
-    TODO: finish me. intrinsic support for C++ is an incomplete feature. Watch
-          https://github.com/UAVCAN/nunavut/issues/91 for progress.
-    {{ composite_type.tag_field_type | declaration }} _tag_;
-    union {
+{#-
+ # Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ # Copyright (C) 2021  UAVCAN Development Team  <uavcan.org>
+ # This software is distributed under the terms of the MIT License.
+#}
+    union // The union is placed first to ensure that the active element address equals the object address.
```
pavel-kirienko, line 6: I am a little uneasy about this implementation because it violates AUTOSAR/MISRA (which prohibit unions) and is generally below the safety bar attainable with sensible C++ code. We spoke about it earlier and I would like to revisit this question now.

Beside the type safety issues, this specific implementation lacks the API handles necessary for constructing any non-trivial template code such as traversing/selecting the options at compile time without hard-coding the available options. This will become important, for instance, when implementing the Register interface, which relies on unions heavily.

We can't use `std::variant` due to our commitment to support old C++ standards, but maybe we can work around this with the help of an option switch that selects which `variant` type to use, similar to what you have for variable-length arrays? If the user wants C++17, then the generated code would simply use `std::variant`. Otherwise, the user would resort to a custom class or one of the several off-the-shelf implementations of the same, like https://github.com/martinmoene/variant-lite (there are others).

This should not increase the implementation workload on you since you will be able to rely on `std::variant` instead of reinventing an alternative type-safe mechanism.

src/nunavut/lang/cpp/support/serialization.j2
```diff
@@ -1,23 +1,67 @@
...
+{%- macro float32_union() -%}
+    typedef union  // NOSONAR
+    {
+        uint32_t bits;
+        {{typename_float_32}} real;
+    } Float32Bits;
+{%- endmacro -%}
+
+// UAVCAN common serialization support routines.                                                             +-+ +-+
+// This file is based on canard_dsdl.h, which is part of Libcanard.                                          | | | |
```
pavel-kirienko, line 24: Is it though

src/nunavut/lang/cpp/support/VariableLengthArray.hpp
```diff
@@ -0,0 +1,464 @@
...
+#include <memory>
+#include <utility>
+
+namespace nunavut
+{
+namespace support
+{
+/**
+ * Minimal, generic container for storing UAVCAN dynamic arrays. One property that is unique
+ * for dynamic arrays is that they have a maximum bound which this implementation enforces.
```
pavel-kirienko, line 23:
```suggestion
 * Minimal, generic container for storing UAVCAN variable-length arrays. One property that is unique
 * for variable-length arrays is that they have a maximum bound which this implementation enforces.
```

"Dynamic" may be misleading because it might be understood as a storage qualifier rather than size qualifier. The Specification never says "dynamic" for this reason, it says "variable-length".

src/nunavut/lang/cpp/support/VariableLengthArray.hpp
```diff
@@ -0,0 +1,464 @@
+/*
+ * Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * Copyright (C) 2014 Pavel Kirienko <pavel.kirienko@gmail.com>
+ * Copyright (C) 2021  UAVCAN Development Team  <uavcan.org>
+ * This software is distributed under the terms of the MIT License.
+ */
```
pavel-kirienko, line 6: Should we not use `//`, `///` for consistency everywhere?

src/nunavut/lang/cpp/support/VariableLengthArray.hpp
```diff
@@ -0,0 +1,464 @@
...
+        }
+
+        return capacity_;
+    }
+
+    /**
+     * Deallocate or reallocate memory such that not more than {@code size()} elements can be stored in this object.
+     *
+     * @return True if nothing was done or if memory was successfully resized. False to indicate that the allocator
+     *         could not provide enough memory to move the existing objects to a smaller allocation.
```
pavel-kirienko, line 237: > False to indicate that the allocator could not provide enough memory to move the existing objects to a smaller allocation

Is there any practical scenario where this information is actionable? How do you envision this method to be used? My understanding is that it may be leveraged in some very specific cases involving non-hard-real-time processes that desire to conserve memory. If the memory could not be reclaimed, from the standpoint of such process it is not an error since it does not affect the business logic (the data is still stored and reachable). If my understanding is correct, then probably every invocation of this method will have to be accompanied with an explicit cast to void, so maybe it should return nothing?

src/nunavut/lang/cpp/support/VariableLengthArray.hpp
```diff
@@ -0,0 +1,464 @@
+/*
```
pavel-kirienko, line 1: Shouldn't it be named `variable_length_array.hpp`?

src/nunavut/lang/cpp/support/VariableLengthArray.hpp
```diff
@@ -0,0 +1,464 @@
...
+    }
+
+    // +----------------------------------------------------------------------+
+    // | MODIFIERS
+    // +----------------------------------------------------------------------+
+    /**
+     * Push a new element on to the back of the array and grow the array size by 1.
+     *
+     * @return A pointer to the stored value or nullptr if there was not enough capacity (use reserve to
+     *         grow the available capacity).
```
pavel-kirienko, line 305: Why can't we have `try_push_back() -> T*` that invokes `reserve()` automatically?

pavel-kirienko (changes requested):
We started this conversation on Slack but it was never completed: can we ship a simple static allocator for use with variable-length arrays and make it the default? The bounded-size nature of DSDL makes it the most natural choice for virtually all cases (with the notable exclusion of #189).

thirtytwobits:
> we ship a simple static allocator for use with variable-length arrays and make it the default?

Yeah. I think that makes sense.

src/nunavut/lang/cpp/support/VariableLengthArray.hpp
thirtytwobits: Because this isn't a generalized container type and I'd never generate code that relied on this. I want code that always calls reserve once to ensure the subsequent logic will be successful.

src/nunavut/lang/cpp/support/VariableLengthArray.hpp
thirtytwobits: bah. Yeah.

src/nunavut/lang/cpp/templates/base.j2
```diff
@@ -0,0 +1,76 @@
...
+-#}
+//
+// This is an AUTO-GENERATED UAVCAN DSDL data type implementation. Curious? See https://uavcan.org.
+// You shouldn't attempt to edit this file.
+//
+// Checking this file under version control is not recommended unless it is used as part of a high-SIL
+// safety-critical codebase. The typical usage scenario is to generate it as part of the build process.
+//
+// To avoid conflicts with definitions given in the source DSDL file, all entities created by the code generator
+// are named with an underscore at the end, like foo_bar_().
```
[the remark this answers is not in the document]
thirtytwobits: Ah, good catch. Copy-and-paste error.

src/nunavut/lang/cpp/support/VariableLengthArray.hpp
pavel-kirienko: Sure but we need to be application-friendly. Say, if I want to push X battery cell voltages into my array, calling `try_push_back()` X times is easier than `reserve()` once and then `push_back_no_alloc()` (both require error checking anyway).

pavel-kirienko:
Also, one thing I forgot to mention: can we please support at least a basic iterator API (pointer-based) for the variable array type? Otherwise, the class would be much less useful when interfacing with 3rd-party code.

src/nunavut/lang/cpp/templates/base.j2
```diff
@@ -0,0 +1,76 @@
+{#-
+ # Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ # Copyright (C) 2021  UAVCAN Development Team  <uavcan.org>
+ # This software is distributed under the terms of the MIT License.
+ # Authors: David Lenfesty, Scott Dixon <dixonsco@amazon.com>, Pavel Kirienko <pavel@uavcan.org>,
+ #          Peter van der Perk <peter.vanderperk@nxp.com>
```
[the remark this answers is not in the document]
thirtytwobits: Sure. Although I don't like putting authors in manually. Git blame does this better.

src/nunavut/lang/cpp/support/VariableLengthArray.hpp
thirtytwobits: An error occurred so it has to be communicated out. If you choose to void cast it away you ignore the error but that is a choice the application has to make not the framework.

thirtytwobits:
> Also, one thing I forgot to mention: can we please support at least a basic iterator API (pointer-based) for the variable array type? Otherwise, the class would be much less useful when interfacing with 3rd-party code.

I'd like to release this initially with the most minimal support needed for ser/des. Let's then analyze usage and expand as real use cases present themselves. I really don't want to provide a generalized container library from Nunavut. That doesn't seem like it's competency.

thirtytwobits:
Also note that the CI will remain broken until we agree on the content of the PR as I'm restructuring the CI interface to better support parallelization going forward.

src/nunavut/lang/cpp/support/serialization.j2
thirtytwobits: I'm not sure I understand your comment here?

src/nunavut/lang/cpp/templates/_fields_as_union.j2
thirtytwobits: I vote for this for C++11 and 14 then switch over to std::variant when generating c++17. Would that be acceptable?

pavel-kirienko: no :) See, if you keep this, there will be two implementations with vastly different APIs. The current one will be mostly useless unless we equip it with compile-time introspection hooks provided by std::variant. I think it is easier to just use a third-party `variant` than reinvent this little wheel.

src/nunavut/lang/cpp/support/serialization.j2
pavel-kirienko: The file does not appear to have much in common with canard_dsdl.h, so maybe the comment is unnecessary. This is not very important though.

src/nunavut/lang/cpp/support/variable_length_array.hpp
```diff
@@ -0,0 +1,464 @@
...
+ * Minimal, generic container for storing UAVCAN variable-length arrays. One property that is unique
+ * for variable-length arrays is that they have a maximum bound which this implementation enforces.
+ * This allows use of an allocator that is backed by statically allocated memory.
+ *
+ * @tparam  T           The type of elements in the array.
+ * @tparam  Allocator   The type of allocator.
+ * @tparam  MaxSize     The maximum allowable size and capacity of the array.
+ */
+template <typename T, typename Allocator, std::size_t MaxSize>
+class VariableLengthArray
```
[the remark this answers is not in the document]
thirtytwobits: Jeebus, how did I miss that? Okay.

src/nunavut/lang/cpp/templates/_fields_as_union.j2
thirtytwobits: poo. I really don't like the amount of Shim code we're introducing here. What 3p variant do you think would work?

pavel-kirienko: How about https://github.com/martinmoene/variant-lite (Boost license)

thirtytwobits: Tracking in #91 TODO list. Let's get this big chunk of incremental work, which is behind an --experimental flag, out.

pavel-kirienko:
@thirtytwobits are you planning to further advance this or do you want it merged as-is?

thirtytwobits:
> @thirtytwobits are you planning to further advance this or do you want it merged as-is?

I have more work to do to respond to all your comments.

thirtytwobits:
@pavel-kirienko , let me know if this is okay to merge as-is (i.e. the c++ work isn't done but we need to get all this change in) and I'll fix up the CI

src/nunavut/lang/cpp/__init__.py
```diff
@@ -345,11 +521,17 @@ def filter_declaration(language: Language,
     if isinstance(instance, pydsdl.PrimitiveType) or isinstance(instance, pydsdl.VoidType):
         return filter_type_from_primitive(language, instance)
     elif isinstance(instance, pydsdl.VariableLengthArrayType):
-        return 'std::vector<{}>'.format(filter_declaration(language, instance.element_type))
+        variable_array_type = language.get_option('variable_array_type', DEFAULT_ARRAY_TYPE)
```
pavel-kirienko, line 524: The `DEFAULT_ARRAY_TYPE` does not seem to be an admissible default here because `std::array` cannot contain size information.

src/nunavut/lang/cpp/support/serialization.j2
```diff
@@ -1,23 +1,67 @@
...
+ #          Peter van der Perk <peter.vanderperk@nxp.com>
+-#}
+
+{%- macro assert(expression) -%}
+    {%- if options.enable_serialization_asserts -%}
+    NUNAVUT_ASSERT({{ expression }});
+    {%- endif -%}
+{%- endmacro -%}
+
+{%- macro float32_union() -%}
```
pavel-kirienko, line 15: This macro will not be needed.

pavel-kirienko (approved):

src/nunavut/lang/cpp/__init__.py
thirtytwobits: https://en.cppreference.com/w/cpp/container/array/size ?

pavel-kirienko: What I mean is a runtime state: a variable-length DSDL array shall be resizable, but `std::array` is not.

thirtytwobits: variable length arrays are vectors in this revision.

pavel-kirienko: That is true only as long as `variable_array_type` is explicitly defined, otherwise it's `std::array`. Maybe this line should rather raise an error if `variable_array_type` is not defined instead of defaulting to the fixed-size array?

PR #199, opened 2021-05-05
Somehow...this works

### LINKED ISSUE #108: Do we care about code generation performance?

thirtytwobits:
Um, well. Colour me embarrassed but it appears that the unique filter does work with the template cache enabled. Paging @PetervdPerk-NXP.

Fixes #108

### DIFF
--- a/src/nunavut/jinja/__init__.py
+++ b/src/nunavut/jinja/__init__.py
@@ -128,7 +128,7 @@ class CodeGenEnvironment(Environment):
                          lstrip_blocks=lstrip_blocks,
                          trim_blocks=trim_blocks,
                          auto_reload=False,
-                         cache_size=0)
+                         cache_size=400)
         self.globals['options'] = _LanguageTemplateNamespace()
         self.globals['ln'] = _LanguageTemplateNamespace()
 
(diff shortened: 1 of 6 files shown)


pavel-kirienko:
@thirtytwobits can you reproduce this problem after enabling the cache? https://github.com/UAVCAN/nunavut/issues/88#issuecomment-546927646

thirtytwobits:
Yep. It works!

pavel-kirienko (approved):

PR #200, opened 2021-06-13
improving coverage and adding language tests

thirtytwobits:

PR #201, opened 2021-06-21
Add documentation generation to Nunavut

bbworld1:
Adds a new `docgen` language to Nunavut, which generates HTML documentation pages for namespaces.

`python3 -m nunavut --experimental-languages --target-language docgen ~/git/sources/public_regulated_data_types/uavcan/` will generate a documentation set in a Namespace.html file.


pavel-kirienko:
Please rebase to main or merge it into your branch. There have been critical changes since April 14.

pavel-kirienko:
I didn't yet look at the code (will do next), only poked around the output. It looks super cool and is almost exactly what I've had in mind while writing the [description on the forum](https://forum.uavcan.org/t/generate-rich-navigable-html-docs-using-nunavut/1163). I noticed an issue and also would like to introduce a few UX suggestions. I am using the latest Firefox for testing.

## Search issues

After you erase the search query, the data type view is not re-populated back:

image

Then, if you navigate to a data type that was not included in the search result, its attributes will not be shown:

image

The usability of the search could be improved perhaps if the matches were highlighted. I am not sure though how difficult it would be to implement that.

## Text display

I think it looks a lot better with Ubuntu Mono compared to the default monospace font used on my OS (Manjaro). Also, maybe we should reduce the line-height from 1.5 to about 1.2 to increase display density? This is what we get then:

image

I suggest increasing the tab width about twofold. The current layout makes nesting a bit difficult to see despite the vertical rulers.

Links to data type definitions seem to stand out a bit too much, maybe we should make them darker?

I would like to make the border between adjacent data types a bit more explicit. Perhaps we could add a pale separation line, similar to the vertical rulers? Or maybe we could make the default body text color slightly grayish by default, turning it to full black on hover, such that you could highlight the type by hovering the mouse over it? The latter wouldn't work with touchscreens though.

## Extent

Extent of composites is shown incorrectly; I think you forgot to divide by 8.

Extent of fields is shown despite the "Show extent" being disabled:

image

## Unions

The semantics of unions is very different from regular structs. Can we please mark unions explicitly?

src/nunavut/cli.py
```diff
@@ -179,13 +184,23 @@ def _build_post_processor_list_from_args(args: argparse.Namespace) \
...
                                allow_overwrite=not args.no_overwrite)
+
+    # If docgen selected, warn about linked namespaces
+    if args.target_language == 'docgen':
+        if len(extra_includes) > 0:
+            print(textwrap.dedent("""
+                W: Other lookup namespaces are linked in these generated docs.
+                If you do not generate docs for these other namespaces as well, links to external
+                data types could be broken (expansion will still work).
+            """).lstrip())
```
pavel-kirienko, line 202: Did you mean to use `logging.warning` here?

src/nunavut/lang/docgen/__init__.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+@template_language_filter(__name__)
+def filter_is_service_req(language: Language, instance: typing.Any) -> bool:
+    return instance.has_parent_service and instance.full_name.split(".")[-1] == "Request"
+
+
+@template_language_filter(__name__)
+def filter_is_field(language: Language, instance: typing.Any) -> bool:
+    return isinstance(instance, pydsdl.Field)
```
pavel-kirienko, line 43: If you rebase to main you can replace these filters with tests (available since #200):

https://github.com/UAVCAN/nunavut/blob/c3c56efec1626c667e087695144882c23642c79a/src/nunavut/lang/__init__.py#L82-L84

src/nunavut/lang/docgen/__init__.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+@template_language_filter(__name__)
+def filter_extent(language: Language, instance: typing.Any) -> str:
+    try:
+        return instance.bit_length_set.max or 0
```
pavel-kirienko, line 49: Extent and the maximum element of the bit length set are not the same thing: https://pydsdl.readthedocs.io/en/stable/pages/pydsdl.html#pydsdl._serializable._composite.DelimitedType

Maybe we actually should display both?

- Extent is useful if you need to determine how much memory is needed to deserialize a data type or how much additional data you can add in its next minor version.
- Max length is useful if you need to determine the worst case network bandwidth utilization by the data type.

pavel-kirienko (changes requested):
@bbworld1 is the root `Namespace.html` self-contained, or does it require the other HTML files to be present as well?

src/nunavut/cli.py
bbworld1: Oh, sorry. I did intend to use `logging.warning` here.

All that warning means is that if the Namespace.html links to other namespaces (i.e. `--lookup`), that namespace needs to be separately generated. Perhaps I can make that wording a bit more clear.

bbworld1:
This should fix all above suggestions.

src/nunavut/cli.py
```diff
@@ -179,13 +185,23 @@ def _build_post_processor_list_from_args(args: argparse.Namespace) \
...
                                allow_overwrite=not args.no_overwrite)
+
+    # If docgen selected, warn about linked namespaces
+    if args.target_language == 'docgen':
+        if len(extra_includes) > 0:
+            logging.warn(textwrap.dedent("""
+                W: Other lookup namespaces are linked in these generated docs.
+                If you do not generate docs for these other namespaces as well, links to external
+                data types could be broken (expansion will still work).
+            """).lstrip())
```
pavel-kirienko, line 203:
```suggestion
            logging.warning(
                "Other lookup namespaces are linked in these generated docs. "
                "If you do not generate docs for these other namespaces as well, "
                "links to external data types could be broken (expansion will still work)."
            )
```

This minor change fixes unidiomatic formatting, removes the unnecessary `W: ` prefix, and replaces the deprecated `warn` with `warning`.

src/nunavut/lang/docgen/templates/Namespace.j2
```diff
@@ -0,0 +1,93 @@
...
+            --bs-font-monospace:
+                "Ubuntu Mono",
+                SFMono-Regular,
+                Menlo,
+                Monaco,
+                Consolas,
+                "Liberation Mono",
+                "Courier New",
+                monospace;
+        }
```
pavel-kirienko, line 45: The Ubuntu font is not selected unless I explicitly load it like

```html
<link rel="preconnect" href="https://fonts.googleapis.com">
<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
<link href="https://fonts.googleapis.com/css2?family=Ubuntu+Mono:ital,wght@0,400;0,700;1,400;1,700&display=swap" rel="stylesheet"> 
```

pavel-kirienko (changes requested):
Natural ordering would help here:

image

There are several suggestions from my prior response that still stand, what do you think of them?

> I suggest increasing the tab width about twofold. The current layout makes nesting a bit difficult to see despite the vertical rulers.
>
>Links to data type definitions seem to stand out a bit too much, maybe we should make them darker?
>
>I would like to make the border between adjacent data types a bit more explicit. Perhaps we could add a pale separation line, similar to the vertical rulers? Or maybe we could make the default body text color slightly grayish by default, turning it to full black on hover, such that you could highlight the type by hovering the mouse over it? The latter wouldn't work with touchscreens though.

Also, one small tweak: would it be difficult to add a button that collapses all branches of the drop-down tree except the selected one? Suppose I was looking at one data type and then I navigated to another one. I don't want the old one to clutter the view so it would be nice to stow it away quickly.

bbworld1:
I think this should address most of what was pointed out; apologies for missing that earlier.

Regarding data type definition links standing out too much, I'm not sure I agree that the color is too distracting. I've removed the underline from the nested data type links. If this is not sufficient we can still try a slightly darker shade.

As for the collapsing of other namespaces, I've just made that the default behavior. There's a checkbox to disable it.

pavel-kirienko (changes requested):
Nice. The comment I left above regarding the sorting of the tree on the left panel still stands. Other than that everything seems to be in order except that we should add a few words to the docs about this feature: the README should be updated and probably something else. @thirtytwobits can you specify please what docs should be updated?

bbworld1:
I think _that_ should resolve all the points @pavel-kirienko pointed out above (except for docs). Apologies for missing that :)

thirtytwobits:
I think `Namespace.html` should be `index.html` instead?

thirtytwobits:
We need "expand all/collapse all" options. Some people like looking at one thing at a time but other (ahem...me) just want a full-text page so I can CMD+f search for things.

pavel-kirienko:
>CMD+f search for things

I am not against this but behold the search bar at the top-left

thirtytwobits:
I should have started with this but: overall, this is freaking fantastic!

thirtytwobits:
> > CMD+f search for things
> 
> I am not against this but behold the search bar at the top-left

That's great too but, for example, if you wanted to find where we used newtons this search would yield nothing but a full text search would show you all use of the word "newtons" (e.g. for uavcan.si.unit.force.Scalar).

src/nunavut/lang/properties.ini
```diff
@@ -277,3 +277,15 @@ named_values =
...
+extension = .html
+has_standard_namespace_files = True
+namespace_is_composite_type = False
+namespace_file_stem = Namespace
+named_values =
+    true = true
+    false = false
+    null = null
+options =
+    omit_serialization_support = true
```
thirtytwobits, line 291: I'm not a fan of this. First off, command line overrides should dominate so if you specify "always" you should override this option (see above). Instead we should make the generator fail gracefully if the language simply doesn't have any serialization support.

src/nunavut/lang/properties.ini
```diff
@@ -277,3 +277,15 @@ named_values =
     true = true
     false = false
     null = null
+
+[nunavut.lang.docgen]
```
thirtytwobits, line 281: I'm thinking this should be `nunavut.lang.html`. We could have a philosophical argument about this but I do believe this is consistent. We are generating a language here but it is not a machine language; it is a natural language and we use HTML to describe it. This would be consistent with other documentation generation targets like `nunavut.lang.pdf`.

thirtytwobits (commented):
src/nunavut/lang/docgen/templates/__init__.py should have at least some comments in it describing what's in the module.

src/nunavut/cli.py
```diff
@@ -15,10 +15,17 @@
...
+import nunavut.lang
 
 
-def _should_generate_support(args: argparse.Namespace) -> bool:
-    if args.generate_support == 'as-needed':
+def _should_generate_support(
+    args: argparse.Namespace,
+    language_context: 'nunavut.lang.LanguageContext'
+) -> bool:
+    if language_context.get_target_language() is not None and \
```
thirtytwobits, line 25: This has mypy errors but I don't think the logic is correct. commandline options should have a higher priority then internal language options

pavel-kirienko:
Here's one extra nit: is it possible to remove the clipping of the main viewing area? As you can see on the screenshot, the main area where the docs are is clipped on the right and at the bottom.

image

bbworld1:
I've replaced the command line override approach with a simpler approach; if Nunavut can't find serialization support for a target language it will just fail gracefully and print a warning for the user. I've also renamed the target to `html` and fixed the clipping issue.

I think that should resolve all that was mentioned above.

src/nunavut/lang/__init__.py
```diff
@@ -231,13 +231,23 @@ def support_files(self) -> typing.Generator[pathlib.Path, None, None]:
...
+            module = importlib.import_module('nunavut.lang.{}.support'.format(self._language_name))
+
+            # All language support modules must provide a list_support_files method
+            # to allow the copy generator access to the packaged support files.
+            list_support_files = getattr(module, 'list_support_files')  \
+                # type: typing.Callable[[], typing.Generator[pathlib.Path, None, None]]
+            return list_support_files()
+        except ModuleNotFoundError:
+            # No serialization support for this language
+            logger.warn("No serialization support for selected target. Skipping serialization support generation.")
```
thirtytwobits, line 244: this is a much better strategy but I'd make this an info message. If the language doesn't have serialization support then that's the nature of that language. There's nothing to warn the user about in such a case.

src/nunavut/cli.py
```diff
@@ -186,6 +186,16 @@ def _build_post_processor_list_from_args(args: argparse.Namespace) \
     if args.generate_support != 'only':
         generator.generate_all(is_dryrun=args.dry_run,
                                allow_overwrite=not args.no_overwrite)
+
+    # If docgen selected, warn about linked namespaces
+    if args.target_language == 'docgen':
```
thirtytwobits, line 191: docgen should be `html` now, right?

thirtytwobits (commented):
almost there. Doing great with this! Thanks so much.

src/nunavut/version.py
```diff
@@ -3,6 +3,6 @@
 # This software is distributed under the terms of the MIT License.
 #
 
-__version__ = "1.2.2"
+__version__ = "1.2.3"
```
pavel-kirienko, line 6:
```suggestion
__version__ = "1.3.0"
```

You probably meant this. The minor version comes after the major version, then comes the patch version.

thirtytwobits (approved):

pavel-kirienko (changes requested):
Can we please update the README also? Currently it says only about C, C++, and Python.

bbworld1:
I've added a section in the README regarding generating html documentation, and bumped the version number to 1.3.0.

pavel-kirienko (approved):

pavel-kirienko:
@thirtytwobits will you press the button, please

thirtytwobits:
Ready to merge. Should I do this now?

pavel-kirienko:
Can't see why not. Do you want @bbworld1 to confirm?

bbworld1:
Let's do it!

PR #202, opened 2021-07-13
Fix JS/CSS Asset Include into Nunavut

bbworld1:
Previously, JS and CSS assets in the HTML template directory would not be included in the built package, causing HTML documentation generation to fail. This PR fixes this issue by including JS and CSS assets via setup.py package_data.


pavel-kirienko (approved):

pavel-kirienko:
@bbworld1 please bump the version number (missed that in my review)

PR #208, opened 2021-08-14
Adding dsdl doc to generated code

thirtytwobits:
Quckie to add your awesome new dsdl docs into the budding C++ types.

### DIFF
--- a/src/nunavut/lang/properties.yaml
+++ b/src/nunavut/lang/properties.yaml
@@ -280,6 +280,27 @@ nunavut.lang.cpp:
     named_values:
         'true': 'true'
         'false': 'false'
+    comment_styles:
+        'zubax':
+            'prefix': ''
+            'comment': '/// '
+            'suffix': ''
+        'cpp':
+            'prefix': ''
+            'comment': '// '
+            'suffix': ''
+        'javadoc':
+            'prefix': '/**'
+            'comment': ' * '
+            'suffix': ' */'
+        'c':
+            'prefix': '/*'
+            'comment': ' * '
+            'suffix': ' */'
+        'qt':
+            'prefix': '/*!'
+            'comment': ' * '
+            'suffix': ' */'
     options:
         target_endianness: any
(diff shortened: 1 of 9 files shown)


src/nunavut/lang/properties.yaml
```diff
@@ -280,6 +280,27 @@ nunavut.lang.cpp:
     named_values:
         'true': 'true'
         'false': 'false'
+    comment_styles:
+        'zubax':
+            'prefix': ''
+            'comment': '/// '
+            'suffix': ''
```
pavel-kirienko, line 287: Shouldn't this one be called either `cpp` or `doxygen`? Documentation comments are usually different from regular comments.

thirtytwobits: `//` is, by convention, referred to as "C++ style" and I've used it here as such. I've only ever seen `///` used for doxygen or by your Zubax style however doxygen style requires "at least two" lines using `///` so if we wanted to call this `cpp-doxygen` we'd need to add `///` as a prefix and a suffix. The problem with doing this is the DSDL comments are not structured so it wouldn't actually result in valid doxygen comments.

pavel-kirienko: >The problem with doing this is the DSDL comments are not structured so it wouldn't actually result in valid doxygen comments.

I don't quite follow, can you elaborate on why pasting DSDL docs may result in invalid Doxygen comments?

pavel-kirienko (approved):

src/nunavut/lang/properties.yaml
thirtytwobits: well, actually, since the DSDL documentation is parsed or stropped there's no guarantee that it is valid documentation for any autodoc system. As such I've changed the names of things around to be about "block comments." If we ever want to parse the docs we can try to transform them into doxygen, sphinx-doc, etc; based on language.

pavel-kirienko (approved):

Issue #209, opened 2021-09-11
Rust Type Generation Support


davidlenfesty:
(In support of the uavcan.rs v1 rewrite)

Rust's type system maps extremely well to UAVCAN's types (struct -> struct, union -> enum), so this should be relatively easy.

In conversation with Alexander Hübener, it was decided we should go with Rust macros to provide serialization, just to homogenize as much as possible to Rust. As well, Rust has excellent testing infrastructure that we can take advantage of, without having to find a way to integrate it into Nunavut's testing.

Definitely not finalized, but this is approximately how I think the generated code should look.

`uavcan/node/Heartbeat.1.0.uavcan`
```rust
// Some namespace collision here, fairly trivial to resolve, but will have to come up with a solution
use uavcan::{Serialize, Deserialize};

#[derive(Debug, Clone, Serialize, Deserialize)]
// Values here are placeholder, no idea what they are, also didn't exhaustively cover attributes
#[uavcan(extent=56)]
pub struct Heartbeat_1_0 {
    #[uavcan(bit_width=32, alignment=1)]
    pub uptime: u32,
    pub health: crate::uavcan::node::Health_1_0,
    pub node: crate::uavcan::node::Mode_1_0,
    #[uavcan(bit_width=8, alignment=1)]
    pub vendor_specific_status_code: u8,
}
```

I'm not entirely sure how the folder structure will have to work out, specifically:
- Will nunavut generate an entire crate?
- Do I need to add `mod.rs` files to each folder to make the hierarchy work out correctly?

pavel-kirienko:
@davidlenfesty do you think we still need the [dsdl_parser.rs](https://github.com/UAVCAN/dsdl_parser.rs), or should its repository be archived?

Nunavut's C generator comes with a rather extensive yet simple test suite that I recommend looking at:

- https://github.com/UAVCAN/nunavut/tree/main/verification/c/suite
- https://github.com/UAVCAN/nunavut/tree/main/.buildkite

[...]

PR #211, opened 2021-09-22
Initial Rust type generation

davidlenfesty:
Working draft of initial Rust type generation. DSDL types map very well to Rust so this part is quite easy. 

This will not include serialization work, however it should not get merged until we have a clear path forward there. See #209 for more discussion there.

Addresses #209.

TODO (in no particular order):
- Support for arrays, etc.
- filter function cleanup (copied from the C impl to start, lots of stuff I don't need and a bit to add)
- Need to figure out if I need to generate crate infrastructure, and if so, implement that.
- cleanup of jinja templates to maintain modularity

### DIFF
--- a/src/nunavut/__init__.py
+++ b/src/nunavut/__init__.py
@@ -113,6 +113,21 @@ class YesNoDefault(enum.Enum):
 
 # +---------------------------------------------------------------------------+
 
+def _camel_to_snake_case(camel: str) -> str:
+    out = ""
+    rd_ptr = 0
+    for i in range(len(camel)):
+        if camel[i].isupper():
+            if i == 0 or camel[i-1].isupper():
+                continue
+
+            out += f"{camel[rd_ptr:i]}_"
+            rd_ptr = i
+
+    out += camel[rd_ptr:]
+
+    return out.lower()
+
 
 class Namespace(pydsdl.Any):
     """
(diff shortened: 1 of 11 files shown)


thirtytwobits (changes requested):
Looks like you have some linter errors that need fixing. See https://nunavut.readthedocs.io/en/latest/docs/dev.html for help running the tox build locally to reproduce.

src/nunavut/__init__.py
```diff
@@ -291,9 +306,16 @@ def __hash__(self) -> int:
     # +-----------------------------------------------------------------------+
     # | PRIVATE
     # +-----------------------------------------------------------------------+
-    def _add_data_type(self, dsdl_type: pydsdl.CompositeType, extension: typing.Optional[str]) -> None:
+    def _add_data_type(self, dsdl_type: pydsdl.CompositeType, extension: typing.Optional[str], snake_case: bool) \
```
davidlenfesty, line 309: This is *extremely* hacky. I'm not really comfortable enough with the codebase to be confident in coming up with a proper idiomatic solution, so if anyone has any ideas, I'm all ears.

For reference, the problem I'm trying to solve:

With the default naming of files, I'll end up with a structure like `uavcan::node::Health_1_0::Health_1_0`, and I *want*
a structure like `uavcan::node::Health_1_0`, without the intermediate module. In order to preserve handy naming of types, I want to rename the intermediate module that contains the type, and the most idiomatic (in Rust anyways) way I can think of is changing the file name to snake case.

davidlenfesty:
My next step is building the crate around the generated code using `support`. I'll have to hack around how it currently writes to output for this to work because I need the following file structure:

```
nunavut_out/
├── build.rs
├── Cargo.toml
└── src
    ├── lib.rs
    └── namespaces
        ├── mod.rs
        ├── my_dsdl_namespace
        │   └── mod.rs
        └── nunavut
            └── mod.rs
```

`nunavut_out/src/namespaces/mod.rs` should be generated by the build script. I've already implemented a proof of concept build script that does what is needed.

src/nunavut/__init__.py
pavel-kirienko: This seems okay and is also exactly the way I handled it in Python.

thirtytwobits: I'm working on a big refactor of the Language and Environment objects because they sort of blew up on me when I tried to add the concept of "using language features" (e.g. `{% ifuses "std_variant" %}...`). This seems co-morbid in that the snake casing probably wants to be shared logic with template filters and other uses.In general, I'd like to get this PR in sooner rather then later which should be fine since we have an "experimental language" flag on by default so users don't see half-baked support like C++ or Rust until we flip that switch. Can we find a good iteration point where we get the tests to run green and we can merge your work with my refactor?

davidlenfesty: Seems like a good idea, but I'm also not entirely sure what you mean in terms of how we manage branches/merging.

The only things I need to get done are a proper solution to snake-casing everything, and the crate generation, basically all the type generation work is done excepting some clean up. Are you suggesting I pull in your refactor now and work with those code changes, or get what I'm doing done and then pull in your refactor?

thirtytwobits: I'm just saying, don't be afraid to iterate in smaller chunks to keep your work from rotting. As long as the build passes and your changes do no harm it's okay to merge them even if incomplete because of the experimental languages flag.

thirtytwobits:
I'm going to close this for now as it's pretty stale. Please resubmit after a rebase if you still want to merge this.

PR #212, opened 2021-10-03
Long-overdue refactor for python core

thirtytwobits:
A lot of stuff got added without a good refactor to DRY out the library
and tighten up the testing. This change  brings HTML into the fold and
prepairs the way for standards variations in templates (e.g. C++11
versus C++17, etc).

The docs were tightened up a bit and the filter/test loading mechanism
was completely rewritten.

A couple of bugs were found and fixed.


pavel-kirienko:
Would be nice to enable pull request builds on ReadTheDocs. Can we do that for Nunavut?

src/nunavut/lang/cpp/templates/_fields_as_variant.j2
```diff
@@ -0,0 +1,36 @@
...
+ # Copyright (C) 2021  UAVCAN Development Team  <uavcan.org>
+ # This software is distributed under the terms of the MIT License.
+#}
+    const std::variant<
+{%- for field in composite_type.fields -%}
+{%- if field is not padding %}
+        {{ field.doc | block_comment('cpp-doxygen', 8, 120) }}
+        {{ field.data_type | declaration }}{% if not loop.last %},{% endif %}
+{%- endif -%}
+{%- endfor %}
```
pavel-kirienko, line 12:
```suggestion
{%- for field in composite_type.fields_except_padding -%}
        {{ field.doc | block_comment('cpp-doxygen', 8, 120) }}
        {{ field.data_type | declaration }}{% if not loop.last %},{% endif %}
{%- endfor %}
```

src/nunavut/lang/cpp/templates/_fields_as_variant.j2
```diff
@@ -0,0 +1,36 @@
...
+    > storage_;
+{%- for field in composite_type.fields -%}
```
pavel-kirienko, line 14: ditto

src/nunavut/lang/cpp/templates/_fields_as_variant.j2
```diff
@@ -0,0 +1,36 @@
...
+{%- if field is not padding %}
+
+    constexpr bool {{ ('is' + field.name.capitalize())  | id }}() const noexcept
+    {
+        return std::holds_alternative<{{ field.data_type | declaration }}>(storage_);
```
pavel-kirienko, line 19: This will break for unions that contain fields of the same type. I suggest either querying `index` or use `std::get_if<std::size_t>`.

src/nunavut/lang/cpp/templates/_fields_as_variant.j2
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    constexpr bool {{ ('get' + field.name.capitalize())  | id }}({{ field.data_type | declaration }}& out) const noexcept
+    {
+        const auto pto = std::get_if<{{ field.data_type | declaration }}>(&storage_);
```
pavel-kirienko, line 24: Same here -- type should be replaced with index.

pavel-kirienko (approved):
- I spy the CI tripped on a feature that was introduced in Python 3.6. Do we really need to keep dragging 3.5 along considering that it's long as EOL?

- The "Python versions supported" property in the manifest is obsolete --- 3.9 is not listed. It tends to always go obsolete so maybe we don't need it, after all, considering that we strive to support all modern versions of Python anyway?

- I noticed a few occurrences of questionable formatting. Are you open to integrating Black?

src/nunavut/lang/cpp/templates/_fields_as_variant.j2
thirtytwobits: oh snap. Thanks!

thirtytwobits:
> Would be nice to enable pull request builds on ReadTheDocs. Can we do that for Nunavut?

That's a thing? I'd love to but I don't know how...

thirtytwobits:
> * I spy the CI tripped on a feature that was introduced in Python 3.6. Do we really need to keep dragging 3.5 along considering that it's long as EOL?

Perhaps but let me query my internal users to see if we're all off of 3.5 for our legacy builds. Frankly, it's not been much of a burden to support. Even random.choices is pretty easy to back-port.

> * The "Python versions supported" property in the manifest is obsolete --- 3.9 is not listed. It tends to always go obsolete so maybe we don't need it, after all, considering that we strive to support all modern versions of Python anyway?

I'll update. I want this to reflect what we test in CI.

> * I noticed a few occurrences of questionable formatting. Are you open to integrating Black?

Yes. This, being a huge PR already, is a great place to do that nuclear option. Let's get your comments addressed first then I'll do one more change that reformats the world.

thirtytwobits:
> Would be nice to enable pull request builds on ReadTheDocs. Can we do that for Nunavut?

Seems like this should be possible (readthedocs/nunavut option for building PRs is checked) but there is no documentation on how to actually hook this up to github. Everything assumes some magical step where "readthedocs" is an approved o-auth application for the UAVCAN organization and there is no documentation on how to set that up.

pavel-kirienko:
> Seems like this should be possible (readthedocs/nunavut option for building PRs is checked) but there is no documentation on how to actually hook this up to github. Everything assumes some magical step where "readthedocs" is an approved o-auth application for the UAVCAN organization and there is no documentation on how to set that up.

Yeah, I just noticed that it's already enabled but not working. I am not sure how to fix it either. It worked for PyUAVCAN and PyDSDL out of the box.

thirtytwobits:
*sigh* 
Looks like RTD won't let me run autoapi before sphinx-build. I'm going to have to revert that part of this change...

thirtytwobits:
What is the `filter_add_uuid` about anyway? It's not a uuid, it's a random sequence of characters. Should we not just use `filter_to_template_unique_name` if we need a template-generation-scope unique identifier?

pavel-kirienko:
>What is the filter_add_uuid about anyway? It's not a uuid, it's a random sequence of characters. Should we not just use filter_to_template_unique_name if we need a template-generation-scope unique identifier?

¯\\\_(ツ)\_/¯ I think this was added by @bbworld1

bbworld1:
My bad, that name should be changed.

PR #213, opened 2021-10-07
Fix #210

### LINKED ISSUE #210: Changing zoom is not feasible in nunaweb docs
Zooming doesn't soft-wrap the text that overflows. This makes zooming useless as it hides some of the content due to it. This also means that **only** people who are happy with the provided font size will use the site. https://nunaweb.uavcan.org/api/storage/docs/docs/uavcan/index.html

silverv:
Fixes #210.


silverv:
I don't know how and I thought separation of issues was preferred, not happy

PR #220, opened 2021-10-23
C++ variant support.

thirtytwobits:
When generating C++14 the built-ins will provide a union of aligned_storage to store union values. When using C++17 this same data member becomes an std::variant.

TODO: need to hook-up C++17 build to CI.


thirtytwobits:
Some design notes: I've purposely avoided too much meta-programming for the C++14 version and provided only the bare-minimum set of APIs needed to actually use the variant. This makes the 2014 code more verifiable and easier to understand even if it is more verbose. By using C++17 you can avoid all Nunavut compatibility wrappers and use the std::variant type directly so the upgrade path is pretty compelling.

src/nunavut/lang/__init__.py
```diff
@@ -673,7 +714,7 @@ def get_output_extension(self) -> str:
             return self._target_language.extension
         else:
             raise RuntimeError(
```
pavel-kirienko, line 716: It would benefit Nunavut to use more structured exception types. For the CLI use case it hardly matters but when integrating Nunavut as a library this becomes important. This is probably not related to this specific changeset but maybe it should be looked at in the future.

src/nunavut/lang/cpp/templates/_fields_as_union.j2
```diff
@@ -3,17 +3,77 @@
...
+            , internal_union_value_()
+        {}
+
+        //
+        // Rule of Five.
+        //
+        VariantType(const VariantType&) = delete;
+        VariantType& operator=(const VariantType&) = delete;
+        VariantType(VariantType&&) = delete;
+        VariantType& operator=(VariantType&&) = delete;
```
pavel-kirienko, line 20: This seems unnecessarily restrictive. Most if not all of the generated types will be POD types that can be safely moved and copied. Removing this possibility pretty much requires the use of dynamic storage for passing objects around the application, which is probably not something we want to require. Shall a non-copyable/non-movable member exist, the default copy/move constructors/operators would trigger a compile-time error anyway, so I don't expect us to lose anything by replacing your `delete` with `default` here.

src/nunavut/lang/cpp/__init__.py
```diff
@@ -549,9 +545,53 @@ def filter_includes(language: Language, t: pydsdl.CompositeType, sort: bool = Tr
...
+    .. invisible-code-block: python
 
-    include_gen = IncludeGenerator(language, t, filter_id, filter_short_reference_name)
-    return include_gen.generate_include_filepart_list(language.extension, sort)
+        from nunavut.lang.cpp import filter_destructor_name
+        from unittest.mock import MagicMock
+        import pydsdl
+
+        my_type = MagicMock(spec=pydsdl.ArrayType)
+        my_type.version = MagicMock()
```
pavel-kirienko, line 563: I don't follow why is this necessary. Copy-paste error? The two references below also seem out of place.

src/nunavut/lang/cpp/__init__.py
```diff
@@ -549,9 +545,53 @@ def filter_includes(language: Language, t: pydsdl.CompositeType, sort: bool = Tr
...
+
+    .. invisible-code-block: python
 
-    include_gen = IncludeGenerator(language, t, filter_id, filter_short_reference_name)
-    return include_gen.generate_include_filepart_list(language.extension, sort)
+        from nunavut.lang.cpp import filter_destructor_name
+        from unittest.mock import MagicMock
+        import pydsdl
+
+        my_type = MagicMock(spec=pydsdl.ArrayType)
```
pavel-kirienko, line 562: This seems fragile because below you clearly assume that this is a FixedLengthArrayType. Did you mean this:

```suggestion
        my_type = MagicMock(spec=pydsdl.FixedLengthArrayType)
```

src/nunavut/lang/cpp/templates/_fields_as_union_specializations.j2
```diff
@@ -0,0 +1,60 @@
...
+    >({{ composite_type | short_reference_name }}::VariantType* v) noexcept
+{
+    {{ field.data_type | declaration }}* result = nullptr;
+    if (v)
+    {
+        {{ composite_type | short_reference_name }}::VariantType& vref = *v;
+        if(vref.tag_ == {{ composite_type | short_reference_name }}::VariantType::IndexOf::{{ field.name | id }})
+        {
+            static_assert(std::is_same<typename {{ composite_type | short_reference_name }}::VariantType::alternative<{{ loop.index0 }}>::type, {{ field.data_type | declaration }}>::value,
+                          "The alternative type determined by the preprocessor disagrees with the generated type.");
```
pavel-kirienko, line 22: I think you meant "compiler" instead of "preprocessor"

src/nunavut/lang/cpp/templates/_fields_as_union.j2
```diff
@@ -3,17 +3,77 @@
...
+        static constexpr const typename VariantType::alternative<I, VariantType>::type* get_if(const VariantType* v) noexcept;
+
+    private:
+
+        void destroy_current()
+        {
+{%- for field in composite_type.fields_except_padding if field is not PrimitiveType %}
+            {% if not loop.first %}else {% endif %}if (tag_ == {{ loop.index0 }})
+            {
+                if (!std::is_trivially_destructible<{{ field.data_type | declaration }}>::value)
```
pavel-kirienko, line 59: What is the purpose of this check? If the destructor is indeed trivial, it should be elided by the compiler anyway.

src/nunavut/lang/cpp/templates/_fields_as_variant.j2
```diff
@@ -3,54 +3,49 @@
  # Copyright (C) 2021  UAVCAN Development Team  <uavcan.org>
  # This software is distributed under the terms of the MIT License.
 #}
-    using {{ composite_type.short_name.capitalize() | id }}VariantType = std::variant<
+    using VariantType = class VariantType_ final : public std::variant<
```
pavel-kirienko, line 6: What is the purpose of this alias? Why not simply `class VariantType final : public std::variant<>`?

src/nunavut/lang/cpp/templates/_fields_as_variant.j2
```diff
@@ -3,54 +3,49 @@
...
+        {
+            IndexOf() = delete;
 {%- for field in composite_type.fields_except_padding %}
+            static constexpr const std::size_t {{ field.name | id }} = {{ loop.index0 }}U;
+{%- endfor %}
+        };
 
-    bool {{ ('is' + field.name.capitalize())  | id }}() const noexcept { return (tag_ == {{ loop.index0 }}); }
+        template<size_t I, typename T>
+        struct alternative;
```
pavel-kirienko, line 24: Should be final

pavel-kirienko (changes requested):

src/nunavut/lang/cpp/templates/_fields_as_variant.j2
thirtytwobits: Yeah, it's pointless. Will change

src/nunavut/lang/cpp/templates/_fields_as_union.j2
thirtytwobits: Yup. Removed

src/nunavut/lang/cpp/templates/_fields_as_union_specializations.j2
thirtytwobits: fixed

src/nunavut/lang/cpp/__init__.py
thirtytwobits: will fix

src/nunavut/lang/cpp/__init__.py
thirtytwobits: Sorry, I'm not getting your comment here. Can you clarify?

src/nunavut/lang/__init__.py
thirtytwobits: Can you open an issue? I suppose you want exceptions that derive from some sort of "nunavut exception"?

src/nunavut/lang/cpp/templates/_fields_as_union.j2
thirtytwobits: Okay. I'll have to write tests to cover this so I'll get to it tomorrow.

src/nunavut/lang/__init__.py
pavel-kirienko: https://github.com/UAVCAN/nunavut/issues/224

src/nunavut/lang/cpp/__init__.py
pavel-kirienko: Neither `ArrayType` nor any of its descendants define an attribute named `version`, which means that the following statements are incorrect and should be removed:

> `my_type.version = MagicMock()`
> `my_type.version.major = 1`
> `my_type.version.minor = 0`

src/nunavut/lang/cpp/templates/_fields_as_union.j2
thirtytwobits: Coded this up. I still beed to write tests and fix the python3.9 CI.

src/nunavut/__init__.py
```diff
@@ -64,55 +64,20 @@
...
+from .lang import LanguageContext
+from .lang._common import IncludeGenerator
 
 if sys.version_info[:2] < (3, 5):  # pragma: no cover
     print("A newer version of Python is required", file=sys.stderr)
     sys.exit(1)
 
 
-@enum.unique
-class YesNoDefault(enum.Enum):
```
pavel-kirienko, line 82: This change breaks the library API. `YesNoDefault` should be re-exported at the top level to avoid breakage.

thirtytwobits: doh! Good catch.

thirtytwobits:
Okay, I think this is all correct now. I just need to fix the CI

pavel-kirienko (approved):
I also tested this against PyUAVCAN's own test suite to ensure the API is compatible, all seems well.

PR #223, opened 2021-12-02
Do not use pkg_resources if importlib.resources is available; fixes #144

### LINKED ISSUE #144: pkg_resources: NotImplementedError: Can't perform this operation for unregistered loader type
I tried upgrading to PyTest v5.3, but Nunavut stopped working when invoked by the testing suite:

```
NotImplementedError("Can't perform this operation for unregistered loader type")
Traceback (most recent call last):

  <...irrelevant...>

  File "/home/travis/build/UAVCAN/pyuavcan/pyuavcan/dsdl/_compiler.py", line 197, in generate_package
    namespace_output_stem='__init__')

  File "/home/travis/virtualenv/python3.7-dev/lib/python3.7/site-packages/nunavut/lang/__init__.py", line 225, in __init__
    self._config = self._load_config(*additional_config_files)

  File "/home/travis/virtualenv/python3.7-dev/lib/python3.7/site-packages/nunavut/lang/__init__.py", line 208, in _load_config
    resources = [r for r in pkg_resources.resource_listdir(__name__, '.') if r.endswith('.ini')]

  File "/home/travis/virtualenv/python3.7-dev/lib/python3.7/site-packages/pkg_resources/__init__.py", line 1164, in resource_listdir
    resource_name

  File "/home/travis/virtualenv/python3.7-dev/lib/python3.7/site-packages/pkg_resources/__init__.py", line 1442, in resource_listdir
    return self._listdir(self._fn(self.module_path, resource_name))

  File "/home/travis/virtualenv/python3.7-dev/lib/python3.7/site-packages/pkg_resources/__init__.py", line 1484, in _listdir
    "Can't perform this operation for unregistered loader type"

NotImplementedError: Can't perform this operation for unregistered loader type
```

It's super weird that the exception comes from `pkg_resources`. It does not seem to be involved here at all.

Nunavut works fine outside of PyTest. I googled around a bit but could not find a quick solution, so I decided to downgrade back to PyTest v4.6 for now.

pavel-kirienko:
Fixes #144

`pkg_resources` are used only if Python is older than 3.7, which will become the oldest supported version of the language in three weeks from now.

The new module `nunavut.utilities` is arranged to be compatible and trivially mergeable with its counterpart from @thirtytwobits WIP branch at #220 

```bash
pip install git+https://github.com/UAVCAN/nunavut@importlib
```


pavel-kirienko:
The tests fail because Nunavut makes invalid assumptions about package resources. It assumes that the path returned by `pkg_resources.get_resource_filename` points inside the package, which is not guaranteed to be the case (same applies to `importlib.resources.path`). Fixing that would require a more invasive refactoring; I understood that right now is not the best time for that due to the imminent merge of your large changeset with C++ codegen.

@thirtytwobits please advise what should be the preferred way forward?

thirtytwobits:
I just pushed a candidate for the variant work. Let's get that PR in and I jump on this issue next

pavel-kirienko:
@thirtytwobits is there anything you would like me to do to help with this?

thirtytwobits:
Looking at this now. Blocked by python 3.10 failure:
```
E   ImportWarning: _SixMetaPathImporter.find_spec() not found; falling back to find_module()
```

thirtytwobits:
Okay. I rebased and force pushed with a change that uses https://github.com/python/importlib_resources. I think this will build but I'm also not sure if I'm using it right. This does add a dependency but we don't have to deal with this problem going forward which I think makes this worthwhile.

thirtytwobits (approved):

PR #232, opened 2021-12-15
Fixing deserialization logic error in C

### LINKED ISSUE #221: C deserialization function returns a false-positive invalid argument error for zero-sized buffers
Here:

https://github.com/UAVCAN/nunavut/blob/a90eac2749c1466bdff71a5bf19085e0c2de2299/src/nunavut/lang/c/templates/deserialization.j2#L14

If `*inout_buffer_size_bytes == 0`, `buffer` may be NULL because there is no data to deserialize, which is not an error.

The check should be updated like:

```diff
-- (buffer == {{ valuetoken_null }})
++ ((buffer == {{ valuetoken_null }}) && (0 != *inout_buffer_size_bytes))
```

This is a problem because libcanard intentionally emits null payload pointers for zero-sized payloads (as documented).

thirtytwobits:
closes #221
Bump version number for release.

### DIFF
--- a/src/nunavut/lang/c/templates/deserialization.j2
+++ b/src/nunavut/lang/c/templates/deserialization.j2
@@ -16,10 +16,9 @@
     {
         return -NUNAVUT_ERROR_INVALID_ARGUMENT;
     }
-    else if (buffer == {{ valuetoken_null }})
+    if (buffer == {{ valuetoken_null }})
     {
-        *inout_buffer_size_bytes = 0;
-        return NUNAVUT_SUCCESS;
+        buffer = (const {{ typename_byte }}*)"";
     }
 {% if t.inner_type.bit_length_set.max > 0 %}
     {{ _deserialize_impl(t) }}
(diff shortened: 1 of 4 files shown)


src/nunavut/lang/c/templates/deserialization.j2
```diff
@@ -11,11 +11,16 @@
...
+    if ((out_obj == {{ valuetoken_null }}) || (inout_buffer_size_bytes == {{ valuetoken_null }}) || {# -#}
+        ((buffer == {{ valuetoken_null }}) && (0 != *inout_buffer_size_bytes)))
     {
         return -NUNAVUT_ERROR_INVALID_ARGUMENT;
     }
+    else if (buffer == {{ valuetoken_null }})
+    {
+        *inout_buffer_size_bytes = 0;
+        return NUNAVUT_SUCCESS;
+    }
```
pavel-kirienko, line 23: This branch will cause invalid deserialization because it bypasses the implicit zero extension rule. Also, there should be an `else` after `else if`. Instead, I suggest this:

```suggestion
    if (buffer == {{ valuetoken_null }})
    {
        buffer = (const {{ typename_byte }}*)"";
    }
```

pavel-kirienko (changes requested):

pavel-kirienko (approved):

pavel-kirienko (approved):

PR #234, opened 2022-01-27
migrate pyuavcan.dsdl generation core to lang/py

coderkalyan:
I patched the nunavut core to support (recursive) directories in the support directory, but then ended up consolidating all the python support code into a single file `nunavut_support.py`. Leaving this here in case it comes in handy in the future, it doesn't add any notable complexity. LMK if it should be removed.

Pyuavcan PR coming up in a minute.


coderkalyan:
Actually, I need to check a couple things. Feel free to review though.

thirtytwobits:
This is great. I'll review this tomorrow

src/nunavut/_utilities.py
```diff
@@ -93,7 +93,9 @@ def test_truth(cls, ynd_value: "YesNoDefault", default_value: bool) -> bool:
     DEFAULT = 2
 
 
-def iter_package_resources(pkg_name: str, *suffix_filters: str) -> Generator[pathlib.Path, None, None]:
+def iter_package_resources(
+        pkg_name: str, 
+        *suffix_filters: str) -> Generator[tuple[pathlib.Path, pathlib.Path], None, None]:
```
pavel-kirienko, line 98: Consider setting up Black locally to avoid complaints from the CI system (it will complain about this fragment). I found autoformatting on-save to be super convenient.

src/nunavut/jinja/__init__.py
```diff
@@ -792,7 +792,7 @@ def __init__(self, namespace: nunavut.Namespace, **kwargs: typing.Any):
     # +-----------------------------------------------------------------------+
     # | AbstractGenerator
     # +-----------------------------------------------------------------------+
-    def get_templates(self) -> typing.Iterable[pathlib.Path]:
+    def get_templates(self) -> typing.Iterable[tuple[pathlib.Path, pathlib.Path]]:
```
pavel-kirienko, line 795: Please add `from future import __annotations__` to avoid breakage with old Pythons. https://www.python.org/dev/peps/pep-0585/

src/nunavut/lang/py/__init__.py
```diff
@@ -332,3 +337,51 @@ def filter_longest_id_length(language: Language, attributes: typing.List[pydsdl.
...
+    if isinstance(t, pydsdl.BooleanType):
+        return "bool"  # TODO: numpy.bool is deprecated in v1.20
+    if isinstance(t, pydsdl.SignedIntegerType):
+        return f"_np_.int{pick_width(t.bit_length)}"
+    if isinstance(t, pydsdl.UnsignedIntegerType):
+        return f"_np_.uint{pick_width(t.bit_length)}"
+    if isinstance(t, pydsdl.FloatType):
+        return f"_np_.float{pick_width(t.bit_length)}"
+    assert not isinstance(t, pydsdl.PrimitiveType), "Forgot to handle some primitive types"
+    return "object"  # TODO: numpy.object is deprecated in v1.20
```
pavel-kirienko, line 366: Indeed it is, this is why the original code uses `bool_` and `object_` instead of `bool` and `object`. What is the point of this change?

Also, the `_np_` prefix is missing for the boolean and primitive cases.

pavel-kirienko (changes requested):
Were there any changes to the Jinja templates?

The most interesting thing I suppose is the test suite. Are you planning to integrate it with the same changeset or should we expect a new one later?

thirtytwobits:
Please run `tox -e lint` locally and correct all formatting and mypy errors.

src/nunavut/lang/py/__init__.py
```diff
@@ -332,3 +337,51 @@ def filter_longest_id_length(language: Language, attributes: typing.List[pydsdl.
...
+
+
+@template_language_filter(__name__)
```
thirtytwobits, line 369: `template_language_filter` must return type `str`. If you want to use the language filter decorator you'll either need to define a new type in templates.py or import `GenericTemplateLanguageFilter` and do:

```
@GenericTemplateLanguageFilter[Sequence[tuple[str, pydsdl.CompositeType]]](__name__)
def filter_newest_minor_version_aliases(
    language: Language, tys: Iterable[pydsdl.CompositeType]
) -> Sequence[tuple[str, pydsdl.CompositeType]]:
    ...
```

thirtytwobits:
You need to add `numpy` to setup.cfg as an install_requires

coderkalyan:
> Please run tox -e lint locally and correct all formatting and mypy errors.

Already done, will push after fixing a couple mypy issues.

src/nunavut/lang/py/__init__.py
coderkalyan: Oh I see, I was checked out on an old commit: this was fixed in e0ec0fbf. I am updating everything now.

PR #238, opened 2022-02-23
Fixing linting error and pytest error

### LINKED ISSUE #237: Fix pytest and mypy errors
Recent changes to upstream dependencies are causing linting and unittesting to fail. Fix.

thirtytwobits:
Pytest 7 and Sybil 3 don't agree. Opened
https://github.com/simplistix/sybil/issues/37 and limited our pytest
dependency to be < 7.0

This fixes #237

### DIFF
--- a/.github/workflows/test_and_release.yml
+++ b/.github/workflows/test_and_release.yml
@@ -101,7 +101,7 @@ jobs:
     needs: test
     strategy:
       matrix:
-        python3-version: ['6', '7', '8', '9']
+        python3-version: ['6', '7', '8', '9', '10']
     steps:
     - uses: actions/checkout@v2
       with:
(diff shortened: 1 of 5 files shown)


.github/workflows/test_and_release.yml
```diff
@@ -101,7 +101,7 @@ jobs:
     needs: test
     strategy:
       matrix:
-        python3-version: ['5', '6', '7', '8']
+        python3-version: ['6', '7', '8', '9']
```
pavel-kirienko, line 104: what about 3.10 btw?

pavel-kirienko (approved):

.github/workflows/test_and_release.yml
thirtytwobits: Huh. it works. Okay. Added.

PR #239, opened 2022-02-23
Fixed windows paths

DNedic:
Fix windows paths that break building on windows, most notably because of static asserts. Backslashes were not properly escaped. In the abscence of a better fix, this replaces backslashes with double backslash in jinja templates, as backslashes should not be used on linux anyway.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -102,7 +102,7 @@ Partial example: generating a C struct
         *
         * Auto-generated, do not edit.
         *
-        * Source file: {{T.source_file_path}}
+        * Source file: {{T.source_file_path | replace("\\", "\\\\")}}
         */
 
         #ifndef {{T.full_name | ln.c.macrofy}}
--- a/docs/templates.rst
+++ b/docs/templates.rst
@@ -397,7 +397,7 @@ given a template ``common_header.j2``::
     *
     * Auto-generated, do not edit.
     *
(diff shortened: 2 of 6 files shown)


thirtytwobits:
Please rebase. I've fixed our CI and upstream dependency issues. Thanks.

README.rst
```diff
@@ -102,7 +102,7 @@ Partial example: generating a C struct
         *
         * Auto-generated, do not edit.
         *
-        * Source file: {{T.source_file_path}}
+        * Source file: {{T.source_file_path | replace("\\", "\\\\")}}
```
thirtytwobits, line 105: We wouldn't need to do this if source_file_path was a `pathlib.Path`, right? Perhaps that's the better fix; to convert this into pathlib?

DNedic: This is a quick fix, the issue has been documented for over a year. There is plenty of time to fix it properly afterwards, i may even give it a shot.

pavel-kirienko: >We wouldn't need to do this if source_file_path was a pathlib.Path, right? Perhaps that's the better fix; to convert this into pathlib?

Check this out: https://github.com/UAVCAN/pydsdl/pull/77

We would have to use `.as_posix()` though to ensure consistent behavior.

thirtytwobits:
Cool. We'll take this and then do #243 later.

PR #248, opened 2022-03-31
Address regression introduced with PyDSDL v1.14

pavel-kirienko:
This was reported this morning by @PonomarevDA

The PyDSDL v1.14 changeset is here: https://github.com/UAVCAN/pydsdl/pull/77

The PyDSDL v1.15 changeset is here: https://github.com/UAVCAN/pydsdl/pull/80

~~I also removed Python 3.6 from the classifiers but this does not address the root of the problem, which is that we have the same information listed in several places, and it is certain that inconsistencies will occasionally emerge. I suggest removing the list of Python versions from the classifiers.~~

### DIFF
--- a/.github/workflows/test_and_release.yml
+++ b/.github/workflows/test_and_release.yml
@@ -101,7 +101,7 @@ jobs:
     needs: test
     strategy:
       matrix:
-        python3-version: ['7', '8', '9', '10']
+        python3-version: ['6', '7', '8', '9', '10']
     steps:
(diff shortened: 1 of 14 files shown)


pavel-kirienko:
@thirtytwobits 

download

thirtytwobits:
I'm not okay with removing Python 3.6 support. Is that a requirement for this change?

pavel-kirienko:
@thirtytwobits support for Python 3.6 is back

PR #252, opened 2022-04-11
Fixing windows

thirtytwobits:
This will enable Nunavut to run on windows which unblocks #60, testing
on Windows.


src/nunavut/cli/__init__.py
```diff
@@ -115,7 +115,7 @@ def _make_parser() -> argparse.ArgumentParser:
 
         Additional directories can also be specified through the environment variable
         UAVCAN_DSDL_INCLUDE_PATH, where the path entries are
```
pavel-kirienko, line 117:
```suggestion
        DSDL_INCLUDE_PATH, where the path entries are
```

I suppose there is no point documenting support for the legacy variable?

pavel-kirienko (approved):

pavel-kirienko:
SonarCloud scan is broken because the project key needs to be updated from `uavcan` to `opencyphal`. Sorry, forgot to notify you about this.

src/nunavut/cli/__init__.py
thirtytwobits: Fixed.

PR #259, opened 2022-07-04
Fix missing array capacity constant generated

ASMfreaK:
This is a fix for @JustinOng from forum.
@pavel-kirienko please, review.

### DIFF
--- a/src/nunavut/lang/cpp/templates/_fields.j2
+++ b/src/nunavut/lang/cpp/templates/_fields.j2
@@ -11,7 +11,7 @@
     {% endif -%}
     {{ field.doc | block_comment('cpp-doxygen', 4, 120) }}
     {{ field.data_type | declaration }} {{ field | id }};
-    {%- if field.data_type is VariableLengthArrayType %}
+    {%- if field.data_type is ArrayType %}
     static constexpr {{ typename_unsigned_length }} {{ field | id }}_ArrayCapacity = {{ field.data_type.capacity }}U;
     {%- endif %}
 {%- endfor -%}
--- a/src/nunavut/lang/cpp/templates/_fields.j2
+++ b/src/nunavut/lang/cpp/templates/_fields.j2
@@ -12,6 +12,6 @@
     {{ field.doc | block_comment('cpp-doxygen', 4, 120) }}
     {{ field.data_type | declaration }} {{ field | id }};
     {%- if field.data_type is VariableLengthArrayType %}
-    static constexpr {{ typename_unsigned_length }} {{ field | id }}_ArrayCapacity = {{ f.data_type.capacity }}U;
+    static constexpr {{ typename_unsigned_length }} {{ field | id }}_ArrayCapacity = {{ field.data_type.capacity }}U;
     {%- endif %}
 {%- endfor -%}
--- a/src/nunavut/lang/cpp/templates/_fields.j2
+++ b/src/nunavut/lang/cpp/templates/_fields.j2
@@ -11,4 +11,7 @@
     {% endif -%}
     {{ field.doc | block_comment('cpp-doxygen', 4, 120) }}
     {{ field.data_type | declaration }} {{ field | id }};
+    {%- if f.data_type is VariableLengthArrayType %}
+    static constexpr {{ typename_unsigned_length }} {{ field | id }}_ArrayCapacity = {{ f.data_type.capacity }}U;
+    {%- endif %}
 {%- endfor -%}
--- a/src/nunavut/lang/cpp/templates/_fields.j2
+++ b/src/nunavut/lang/cpp/templates/_fields.j2
@@ -11,7 +11,7 @@
     {% endif -%}
     {{ field.doc | block_comment('cpp-doxygen', 4, 120) }}
     {{ field.data_type | declaration }} {{ field | id }};
-    {%- if f.data_type is VariableLengthArrayType %}
+    {%- if field.data_type is VariableLengthArrayType %}
     static constexpr {{ typename_unsigned_length }} {{ field | id }}_ArrayCapacity = {{ f.data_type.capacity }}U;
     {%- endif %}
 {%- endfor -%}


JustinOng:
Thank you (again) for the very prompt response!

I tried out the patch and got an error `nunavut.jinja.jinja2.exceptions.UndefinedError: 'f' is undefined`.

Replacing `f` with `field` as per below, I can confirm this resolves the initial issue.
```
{%- if field.data_type is VariableLengthArrayType %}
static constexpr {{ typename_unsigned_length }} {{ field | id }}_ArrayCapacity = {{ field.data_type.capacity }}U;
```

pavel-kirienko (approved):

src/nunavut/lang/cpp/templates/_fields.j2
```diff
@@ -11,4 +11,7 @@
     {% endif -%}
     {{ field.doc | block_comment('cpp-doxygen', 4, 120) }}
     {{ field.data_type | declaration }} {{ field | id }};
+    {%- if field.data_type is VariableLengthArrayType %}
```
pavel-kirienko, line 14:
```suggestion
    {%- if field.data_type is ArrayType %}
```

On second thought, there is no point specializing this for variable-length array types only, I think.

ASMfreaK: I think, there is value in consistency in providing this constant for fixed-sized arrays, yes.

src/nunavut/lang/cpp/templates/_fields.j2
```diff
@@ -11,4 +11,7 @@
     {% endif -%}
     {{ field.doc | block_comment('cpp-doxygen', 4, 120) }}
     {{ field.data_type | declaration }} {{ field | id }};
+    {%- if field.data_type is ArrayType %}
+    static constexpr {{ typename_unsigned_length }} {{ field | id }}_ArrayCapacity = {{ field.data_type.capacity }}U;
```
thirtytwobits, line 15: When programming C++ we should not degrade to C. This is a C-like solution and isn't consistent with using the rest of the C++ types. How about we define something like this:
```
template <typename T, std::size_t MaxSize, typename Allocator = std::allocator<T>>
class VariableLengthArrayAsVector : public std::vector<T, Allocator>
{
public:
    VariableLengthArrayAsVector(std::initializer_list<T> l) :
       std::vector<T, Allocator>(l)
    {}

    static constexpr const std::size_t max_size = MaxSize;
};
```

While this has the advantage of allowing consumers to treat VariableLengthArrayAsVector types as a plan `std::vector` it does mean we have to adapt our baseclass constructors and template parameters to be in-sync with vector (for example we'd need to change the default allocator to `std::pmr::polymorphic_allocator` if generating c++17 or above).

thirtytwobits (changes requested):

src/nunavut/lang/cpp/templates/_fields.j2
thirtytwobits: Another, similar option is to generate a type for each of these fields:
```
struct FooType
{
    struct BarType : public std::vector<std::uint8_t> {
        static constexpr const std::size_t max_size = 4;
    } bar;
};
```

Where the type is \<Field Name\> + "Type"

ASMfreaK: Do we really need to implement all the constructors? Can't we say

```cpp
template <typename T, std::size_t MaxSize, typename Allocator = std::allocator<T>>
class VariableLengthArrayAsVector : public std::vector<T, Allocator>
{
public:
    using std::vector<T, Allocator>::vector;

    static constexpr const std::size_t max_size = MaxSize;
};
```

So all the `std::vector` constructors will be inherited?

I like this approach. Maybe we can make this pluggable? So the user can choose which VLA type they want from either a predefined list or a custom value?
As of right now, there are several variants/:

1. plain `std::vector` w/o any max size information
2. support library type based on `std::vector`
3. custom support library type
4. external type

ASMfreaK: I think we can add/modify the following options:

* `variable_lenght_array_template` - a `str`, defaults to `std::vector`
* `variable_lenght_array_template_header` - header name, defaults to `<vector>`
* `variable_lenght_array_extra_template_arguments` - a list of str, defaults to empty list

So the full type would be:
```jinja

{{ variable_lenght_array_template}<{field| type}}{%for arg in variable_lenght_array_extra_template_arguments%},{arg}{% endif %}>
```

thirtytwobits: See https://github.com/OpenCyphal/nunavut/blob/main/src/nunavut/lang/properties.yaml#L315 . My hope is we can tweak this property to do what we need.

thirtytwobits:
ping, can you do the final changes requested for this PR @ASMfreaK ?

src/nunavut/lang/cpp/templates/_fields.j2
thirtytwobits: > Do we have any use cases in mind where such reconfigurability would actually be useful? Or is it a case of YAGNI?

Given that variable length arrays (and strings) is a significant source of pain for embedded systems I think providing a couple of bad defaults (i.e. vector using dynamic memory and VariableLengthArray being a barely serviceable but deterministic implementation) requires us to also allow the user to provide a good override that is optimal for their system.

thirtytwobits:
This is blocking #262 from working so I'm going to take it over.

thirtytwobits:
I'm going to do this:

1. `variable_array_type_template` and `variable_array_type_include` properties will be looked for and consumed but will not have a default value. Instead we'll generate `variable_length_array.hpp` types by default from now on.
2. I'll document (and test) how to use std::vector by defining a wrapper type that supports the `variable_array_type` concept and using the aforementioned language properties. This will also allow any container type supporting the `variable_array_type` concept to be substituted using the same technique.

@pavel-kirienko , any opinions on that plan?

pavel-kirienko:
@thirtytwobits  I have no objections.

thirtytwobits:
Closing this PR. Will post new PR based on this one shortly...

PR #263, opened 2022-08-18
Patch 1

thirtytwobits:
As promised, completed work to support the built-in variable_length_array built-in.

I still owe this project:
1. long-form guide on all these options we're enabling for C and C++
2. tests covering user-supplied overrides of the built-in VLA type.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -14,6 +14,7 @@ import subprocess
 import tempfile
 import textwrap
 import typing
+import urllib
 from doctest import ELLIPSIS
 
 import pydsdl
@@ -187,7 +188,7 @@ class GenTestPaths:
         The directory to place test output under for this test case.
         """
         if self._out_dir is None:
-            self._out_dir = self.create_new_temp_dir(self.test_name)
+            self._out_dir = self.create_new_temp_dir(urllib.parse.quote_plus(self.test_name))
         return self._out_dir
 
     @property
--- a/setup.cfg
+++ b/setup.cfg
@@ -7,7 +7,7 @@ description = Generate code from DSDL using Jinja2 templates.
 long_description = file: README.rst
 long_description_content_type = text/x-rst
 license = MIT
-license_file = LICENSE.rst
+license_files = LICENSE.rst
 keywords = uavcan, dsdl, can, can-bus, codegen, cyphal, opencyphal
 classifiers =
     Development Status :: 3 - Alpha
--- a/src/nunavut/_utilities.py
+++ b/src/nunavut/_utilities.py
@@ -93,6 +93,18 @@ class YesNoDefault(enum.Enum):
     DEFAULT = 2
 
 
+@enum.unique
+class ResourceType(enum.Enum):
+    """
+    Common Nunavut classifications for Python package resources.
+    """
+
+    ANY = 0
+    CONFIGURATION = 1
+    SERIALIZATION_SUPPORT = 2
+    TYPE_SUPPORT = 3
+
+
 def iter_package_resources(pkg_name: str, *suffix_filters: str) -> Generator[pathlib.Path, None, None]:
     """
     >>> from nunavut._utilities import iter_package_resources
@@ -104,5 +116,22 @@ def iter_package_resources(pkg_name: str, *suffix_filters: str) -> Generator[pat
 
     """
     for resource in importlib_resources.files(pkg_name).iterdir():
-        if any(suffix == resource.suffix for suffix in suffix_filters):  # type: ignore
-            yield cast(pathlib.Path, resource)
+        if resource.is_file() and isinstance(resource, pathlib.Path):
+            # Not sure why this works but it's seemed to so far. importlib_resources.as_file(resource)
+            # may be more correct but this can create temporary files which would disappear after the iterator
+            # had copied their paths. If you are reading this because this method isn't working for some packaging
+            # scheme then we may need to use importlib_resources.as_file(resource) to create a runtime cache of
+            # temporary objects that live for a given nunavut session. This, of course, wouldn't help across sessions
+            # which is a common use case when integrating Nunavut with build systems. So...here be dragons.
+            file_resource = cast(pathlib.Path, resource)
+            if any(suffix == file_resource.suffix for suffix in suffix_filters):
+                yield file_resource
+
+
+def empty_list_support_files() -> Generator[pathlib.Path, None, None]:
+    """
+    Helper for implementing the list_support_files method in language support packages. This provides an empty
+    iterator with the correct type annotations.
+    """
+    # works in Python 3.3 and newer. Thanks https://stackoverflow.com/a/13243870
+    yield from ()
--- a/src/nunavut/jinja/__init__.py
+++ b/src/nunavut/jinja/__init__.py
@@ -19,7 +19,7 @@ import nunavut.generators
 import nunavut.lang
 import nunavut.postprocessors
 import pydsdl
-from nunavut._utilities import YesNoDefault
+from nunavut._utilities import ResourceType, YesNoDefault
 from yaml import Dumper as YamlDumper
 from yaml import dump as yaml_dump
 
@@ -779,9 +779,11 @@ class SupportGenerator(CodeGenerator):
         target_language = self.language_context.get_target_language()
 
         self._sub_folders = None  # type: typing.Optional[pathlib.Path]
-        self._support_enabled = False  # If not enabled then we remove any support files found
+        self._serialization_support_enabled = (
+            False  # If not enabled then we remove any serialization support files found
+        )
         if target_language is not None:
-            self._support_enabled = not target_language.omit_serialization_support
+            self._serialization_support_enabled = not target_language.omit_serialization_support
 
             #  Create the sub-folder to copy-to based on the support namespace.
             self._sub_folders = pathlib.Path("")
@@ -794,11 +796,11 @@ class SupportGenerator(CodeGenerator):
     # +-----------------------------------------------------------------------+
     def get_templates(self) -> typing.Iterable[pathlib.Path]:
         files = []
-        target_language = self.language_context.get_target_language()
-
-        if target_language is not None:
-            for resource in target_language.support_files:
+        if self._serialization_support_enabled:
+            for resource in self._get_templates_by_support_type(ResourceType.SERIALIZATION_SUPPORT):
                 files.append(resource)
+        for resource in self._get_templates_by_support_type(ResourceType.TYPE_SUPPORT):
+            files.append(resource)
         return files
 
     def generate_all(self, is_dryrun: bool = False, allow_overwrite: bool = True) -> typing.Iterable[pathlib.Path]:
@@ -812,6 +814,14 @@ class SupportGenerator(CodeGenerator):
     # +-----------------------------------------------------------------------+
     # | Private
     # +-----------------------------------------------------------------------+
+    def _get_templates_by_support_type(self, resource_type: ResourceType) -> typing.Iterable[pathlib.Path]:
+        files = []
+        target_language = self.language_context.get_target_language()
+
+        if target_language is not None:
+            for resource in target_language.get_support_files(resource_type):
+                files.append(resource)
+        return files
 
     def _generate_all(
         self, target_language: nunavut.lang.Language, sub_folders: pathlib.Path, is_dryrun: bool, allow_overwrite: bool
@@ -833,26 +843,13 @@ class SupportGenerator(CodeGenerator):
         for resource in self.get_templates():
             target = (target_path / resource.name).with_suffix(target_language.extension)
             logger.info("Generating support file: %s", target)
-            if not self._support_enabled:
-                self._remove_header(target, is_dryrun, allow_overwrite)
-            elif resource.suffix == TEMPLATE_SUFFIX:
+            if resource.suffix == TEMPLATE_SUFFIX:
                 self._generate_header(resource, target, is_dryrun, allow_overwrite)
                 generated.append(target)
             else:
                 self._copy_header(resource, target, is_dryrun, allow_overwrite, line_pps, file_pps)
                 generated.append(target)
         return generated
-
-    def _remove_header(self, target: pathlib.Path, is_dryrun: bool, allow_overwrite: bool) -> None:
-        if not is_dryrun:
-            if not allow_overwrite and target.exists():
-                raise PermissionError("{} exists. Refusing to remove.".format(str(target)))
-            try:
-                target.unlink()
-            except FileNotFoundError:
-                # missing_ok was added in python 3.8 so this try/except statement will
-                # go away someday when python 3.7 support is dropped.
-                pass
 
     def _generate_header(
         self, template_path: pathlib.Path, output_path: pathlib.Path, is_dryrun: bool, allow_overwrite: bool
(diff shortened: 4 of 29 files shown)


conftest.py
```diff
@@ -187,7 +188,7 @@ def out_dir(self) -> pathlib.Path:
         The directory to place test output under for this test case.
         """
         if self._out_dir is None:
-            self._out_dir = self.create_new_temp_dir(self.test_name)
+            self._out_dir = self.create_new_temp_dir(urllib.parse.quote_plus(self.test_name))
```
thirtytwobits, line 191: To make our unittests Windows safe.

setup.cfg
```diff
@@ -7,7 +7,7 @@ description = Generate code from DSDL using Jinja2 templates.
 long_description = file: README.rst
 long_description_content_type = text/x-rst
 license = MIT
-license_file = LICENSE.rst
+license_files = LICENSE.rst
```
thirtytwobits, line 10: a linter told me to do this.

src/nunavut/_utilities.py
```diff
@@ -93,6 +93,18 @@ def test_truth(cls, ynd_value: "YesNoDefault", default_value: bool) -> bool:
     DEFAULT = 2
 
 
+@enum.unique
+class ResourceType(enum.Enum):
```
thirtytwobits, line 97: This is part of the change to differentiate between support code that is only for serialization and support code that is required for generated PODs

src/nunavut/jinja/__init__.py
```diff
@@ -779,9 +779,11 @@ def __init__(self, namespace: nunavut.Namespace, **kwargs: typing.Any):
         target_language = self.language_context.get_target_language()
 
         self._sub_folders = None  # type: typing.Optional[pathlib.Path]
-        self._support_enabled = False  # If not enabled then we remove any support files found
+        self._serialization_support_enabled = (
+            False  # If not enabled then we remove any serialization support files found
+        )
         if target_language is not None:
-            self._support_enabled = not target_language.omit_serialization_support
+            self._serialization_support_enabled = not target_language.omit_serialization_support
```
thirtytwobits, line 786: previously we only has support files that were for serialization so we'd omit generating them if serialization was turned off. This change allows us to keep disabling serialization support generation but always generate support types used by generated objects.

src/nunavut/lang/__init__.py
```diff
@@ -18,8 +18,8 @@
 
 import pydsdl
 
+from .._utilities import ResourceType, YesNoDefault, iter_package_resources, empty_list_support_files
```
thirtytwobits, line 21: more changes in this py file for support file generation

src/nunavut/lang/_common.py
```diff
@@ -14,6 +14,7 @@
 import typing
 
 import pydsdl
+from nunavut._utilities import ResourceType
```
thirtytwobits, line 17: Providing the right include statements based on support files generated

src/nunavut/lang/c/support/__init__.py
```diff
@@ -8,15 +8,17 @@
 """
 import pathlib
 import typing
-from nunavut._utilities import iter_package_resources
+
```
thirtytwobits, line 11: more support file changes

src/nunavut/lang/cpp/__init__.py
```diff
@@ -8,25 +8,24 @@
     module will be available in the template's global namespace as ``cpp``.
 """
 
-import functools
 import fractions
+import functools
```
thirtytwobits, line 12: in the CPP changes we have both support file generation changes but we also move away from `std::vector` as the default variable-length array and, instead, use our built-in VariableLengthArray template. More on the design of this below.

The intent is to allow, if not encourage, users to provide their own VariableLengthArray implementation or their own allocator for the built-in version. That will be added in a subsequent change.

src/nunavut/lang/cpp/support/__init__.py
```diff
@@ -10,13 +10,13 @@
 import pathlib
 import typing
 
-from nunavut._utilities import iter_package_resources
+from nunavut._utilities import ResourceType, empty_list_support_files, iter_package_resources
```
thirtytwobits, line 13: more support file generation changes

src/nunavut/lang/cpp/templates/deserialization.j2
```diff
@@ -183,11 +183,11 @@
         {% set ref_size = 'size'|to_template_unique_name %}
         // Array length prefix: {{ t.length_field_type }}
         {{ _deserialize_integer(t.length_field_type, ('const %s %s'|format((t.length_field_type | declaration), ref_size)) , offset) }}
-        if ( {{ ref_size}} > {{ t.capacity }}U)
+        if ( {{ ref_size }} > {{ t.capacity }}U)
```
thirtytwobits, line 186: changes to the serialization logic to use reserve which is lighter-weight since no object initialization occurs within this call.

src/nunavut/lang/cpp/templates/deserialization.j2
```diff
@@ -199,8 +199,11 @@
 {% endif %}
     {# GENERAL CASE #}
     {% set ref_index = 'index'|to_template_unique_name %}
-        for (size_t {{ ref_index }} = 0U; {{ ref_index }} < {{ reference }}.size(); ++{{ ref_index }})
+        for (size_t {{ ref_index }} = 0U; {{ ref_index }} < {{ ref_size }}; ++{{ ref_index }})
         {
+            // TODO This is terribly inefficient. We need to completely refactor this template to use C++ emplace and
```
thirtytwobits, line 204: This is crap but part of the "experimental" state of C++ generation. We'll fix this before calling it GA.

src/nunavut/lang/properties.yaml
```diff
@@ -312,7 +312,9 @@ nunavut.lang.cpp:
         enable_serialization_asserts: false
         enable_override_variable_array_capacity: false
         std: c++14
-        variable_array_type: "std::vector<{TYPE},std::allocator<{TYPE}>>"
+        # Provide non-empty values to override the type used for variable-length arrays in C++ types.
+        variable_array_type_template: ""
```
thirtytwobits, line 316: The internal logic is now "if user doesn't override then use our built-in VLA type."

tox.ini
```diff
@@ -249,3 +250,10 @@ deps =
     {[dev]deps}
     {[testenv:docs]deps}
     {[testenv:lint]deps}
+commands =
+    mypy -m nunavut \
```
thirtytwobits, line 254: This change pulls some missing linting dependencies for the `tox -e local` target.

verification/CMakeLists.txt
```diff
@@ -214,16 +216,11 @@ create_dsdl_target(dsdl-test
...
-# | FLAG SETS
-# +---------------------------------------------------------------------------+
-string(TOLOWER ${NUNAVUT_VERIFICATION_LANG} LOCAL_VERIFICATION_LANG)
-
-if (LOCAL_VERIFICATION_LANG STREQUAL "c")
-     set(LOCAL_VERIFICATION_LANG_STANDARD_C "${NUNAVUT_VERIFICATION_LANG_STANDARD}")
-     set(LOCAL_VERIFICATION_LANG_STANDARD_CXX "")
-     list(APPEND LOCAL_ADDITIONAL_DSDL_LIBS "")
-elseif(LOCAL_VERIFICATION_LANG STREQUAL "cpp")
+if(LOCAL_VERIFICATION_LANG STREQUAL "cpp")
```
thirtytwobits, line 219: Just refactoring in this file. No logic change. I wanted to separate out the section that determined compiler flags from the section that added in c-type support for the c++ tests.

src/nunavut/lang/cpp/support/variable_length_array.hpp
```diff
@@ -9,25 +9,56 @@
...
 namespace nunavut
 {
 namespace support
 {
+
+/**
+ * Default allocator using malloc and free.
+ */
+template <typename T>
+class MallocAllocator
```
thirtytwobits, line 31: We're driving towards no malloc or free in the default generated c++ but we're not quite there yet. This is an improvement from relying on std::vector, however, since that's a pretty deep dependency into the STL including a use of the default `Allocator`.

Besides malloc and free, we currently use a lot of static/compile-time support from the standard library, `std::array`, and (if generating c++17) `std::variant`. Neither variant nor array allocate memory internally.

src/nunavut/lang/cpp/support/variable_length_array.hpp
```diff
@@ -39,47 +70,274 @@ class VariableLengthArray
         , alloc_()
     {}
 
-    explicit VariableLengthArray(const Allocator& alloc) noexcept(std::is_nothrow_copy_constructible<Allocator>::value)
+    VariableLengthArray(std::initializer_list<T> l) noexcept
```
thirtytwobits, line 73: The design of the built-in VLA walks a fine line between utility and avoiding over-generalization. That is, we're not Boost; we're not trying to provide a great VLA type you should use everywhere. Instead, I'm trying to provide just enough support that the default c++ types are usable and that are optimized for our serialization logic. For example, if you used this type in generic code you might get pretty frustrated by `push_back` not resizing the array when you are up against the current capacity. There'd be a lot of `if at capacity, grow, then push_back` logic polluting your beautiful code. For Nunavut, you are only consuming the output of our deserializers so you wouldn't be exposed to this wart. Internally, this allows us to remove a lot of logic from push_back since our deserializers tend to optimize themselves in ways that allow a single `reserve` call.

This design philosophy means we don't provide a lot of support for going to and from STL containers. The copy and move constructors, for example, do not resolve to any container supporting the internal concepts. Instead we only allow copy, move, compare between `VariableLengthArray` object and only if they are exactly the same Type, MaxSize, and Allocator.

src/nunavut/lang/cpp/support/variable_length_array.hpp
```diff
@@ -355,6 +640,65 @@ class VariableLengthArray
...
+        internal_compare_element(const U& lhs, const U& rhs, typename std::enable_if<!std::is_floating_point<U>::value>::type* = 0) noexcept
+    {
+        return (lhs == rhs);
+    }
+
+    template <typename U>
+    static constexpr bool
+        internal_compare_element(const U& lhs, const U& rhs, typename std::enable_if<std::is_floating_point<U>::value>::type* = 0) noexcept
+    {
+        // From the C++ documentation for std::numeric_limits<T>::epsilon()
```
thirtytwobits, line 654: I'm less sure about this choice. I needed to provide some sort of floating-point comparison specialization or the container would not support comparisons (I think std::variant requires this support?). My implementation is correct from an IEEE-754 standpoint (I think - this is why you are reviewing) but it is not correct if the units represented by different floating point data is different.

I wonder if we need an additional template parameter for the VLA type that represents units (ULP, perhaps?) such that two VLAs that happened to be the same floating point type and the same max size could not be compared using the built-in comparator?

thirtytwobits (commented):
Adding comments to guide the reviewer.

src/nunavut/lang/cpp/support/variable_length_array.hpp
pavel-kirienko: >I think std::variant requires this support?

You can instantiate `std::variant<Ts...>` even for Ts that do not define comparison operators. Given that, do we really want to define an opinionated default comparison for VLA?

>I wonder if we need an additional template parameter for the VLA type that represents units (ULP, perhaps?)

Why not simply allow the user to define a free-standing spaceship operator arbitrarily?

src/nunavut/lang/cpp/support/variable_length_array.hpp
pavel-kirienko: Given that the capacity (the maximum size) is known at compile time and that we are moving towards a more static design, do we not want to call `reserve(MaxSize)` in the default constructor so that the user doesn't have to deal with this manually afterward? If this default behavior is undesirable we could provide a constructor overload that takes a tag type that suppresses this behavior or simply takes the desired size of the array like std::vector does.

verification/cpp/suite/test_unionant.cpp
```diff
@@ -116,13 +116,21 @@ TEST(UnionantTests, union_value_copy_ctor)
...
     ASSERT_NE(nullptr, ValueType::VariantType::get_if<ValueType::VariantType::IndexOf::empty>(&a.union_value));
     uavcan::primitive::String_1_0& a_result = a.union_value.emplace<ValueType::VariantType::IndexOf::string>(
-        uavcan::primitive::String_1_0{std::vector<std::uint8_t>(hello_world.begin(), hello_world.end())});
+        uavcan::primitive::String_1_0
+        {
+            {
+                reinterpret_cast<const unsigned char*>(hello_world),
+                reinterpret_cast<const unsigned char*>(&hello_world[11]),
+                11
+            }
```
pavel-kirienko, line 131: Speaking of strings, perhaps we should prioritize this: https://github.com/OpenCyphal/specification/issues/51 ? The confusion between arrays of numbers, strings, and arrays of bytes is getting in the way occasionally.

src/nunavut/lang/cpp/support/variable_length_array.hpp
thirtytwobits: No. I considered this design but then why have a VLA type at all and not just use arrays with a size property. The end goal is to allow some sort of allocation to occur that means a constrained environment doesn't need to keep the max size of all variable length array types it deserializes allocated.

src/nunavut/lang/cpp/support/variable_length_array.hpp
thirtytwobits: I went back and forth on this. Without this support no generated C++ type that uses or depends on a VLA type can be compared. Is that acceptable? I'm not quite sure.

SchoberMJ (commented):
Overall looks good, not a C++ guru so I'm not sure I can provide appropriate feedback to the internal_compare function of the VLA.

Python code gen sections look clean. I'm not familiar with the documentation format of the doc strings, in particular lines 134-154 of src/nunavut/lang/cpp/__init__.py just looks like left over code to me.

src/nunavut/jinja/__init__.py
```diff
@@ -779,9 +779,11 @@ def __init__(self, namespace: nunavut.Namespace, **kwargs: typing.Any):
         target_language = self.language_context.get_target_language()
 
         self._sub_folders = None  # type: typing.Optional[pathlib.Path]
-        self._support_enabled = False  # If not enabled then we remove any support files found
+        self._serialization_support_enabled = (
+            False  # If not enabled then we remove any serialization support files found
```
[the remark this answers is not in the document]
thirtytwobits: This is a vestigial comment. I'll remove; thanks.

src/nunavut/lang/cpp/support/variable_length_array.hpp
pavel-kirienko: Okay, but I am uneasy about sacrificing usability for performance. Can you please elaborate on why this is not a case of premature optimization? If we were to embed a call to `reserve()` from `push_back()` with some standard resize policy (e.g., golden ratio), what would be the actual downside aside from the marginal increase in complexity?

>For Nunavut, you are only consuming the output of our deserializers so you wouldn't be exposed to this wart.

You would be exposed to it during serialization and also if DSDL objects make their way into the business logic of the application.

pavel-kirienko (approved):

pavel-kirienko:
There is one potentially serious issue discovered by Sonar though: https://sonarcloud.io/project/issues?pullRequest=263&issues=AYKyzbSpjX57U6EJSHck&open=AYKyzbSpjX57U6EJSHck&id=OpenCyphal_nunavut

PR #265, opened 2022-11-03
First iteration on v2

thirtytwobits:
Nunavut V2 is not a huge change from v1 but it the Python APIs are different enough to warrent the semantic, major revision bump.

The core difference is a refactor of the Language objects to fully support the properties.yaml schema and allow overrides to be provided to nnvg.

A follow-up to this change, before v2 is published, will be a refactor of the internal package and module names to improve encapsulation and to be more consistent with the style of other OpenCyphal python projects.

### DIFF
--- a/.github/workflows/test_and_release.yml
+++ b/.github/workflows/test_and_release.yml
@@ -164,6 +164,25 @@ jobs:
                           --endianness ${{ matrix.endianness }} \
                           ${{ matrix.flag }}
 
+  language-verification-cpp-11:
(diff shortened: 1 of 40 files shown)


pavel-kirienko (commented):
I don't have much to add here, although I must admit I did not look deep. If you want me to pay attention to something specific, let me know. One thing that stands out is that this changeset does not update the documentation, which is odd.

pavel-kirienko:
I think I might have overlooked things; disregard my comment about the docs, please.

PR #266, opened 2022-11-05
refactoring Python API

thirtytwobits:
Using Pavel's pydsdl export strategy to limit the Nunavut API surface


thirtytwobits:
Note that this is all renaming and changing to the doc index. There are no functional changes in this commit.

pavel-kirienko (approved):

PR #268, opened 2022-11-07
Add guard to try / catch inside variable length arrays, allow compilation with -fno-exceptions

Germwalker:
Hello, 

That's it. Use **-fno-exceptions** with the C++ generated variable_length_array.hpp header. 
I need this flag as it get back a lot of heap space.

Maybe a better logic could be implemented like just returning 0 if the vector is already full ?

### DIFF
--- a/src/nunavut/lang/cpp/support/variable_length_array.hpp
+++ b/src/nunavut/lang/cpp/support/variable_length_array.hpp
@@ -468,7 +468,7 @@ public:
             (void) e;
         }
 #else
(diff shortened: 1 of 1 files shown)


thirtytwobits:
Can you add these two changes to fix the linter? https://github.com/OpenCyphal/nunavut/commit/9ed45f8653631548cd942db00f8d116d2f30cd89

thirtytwobits:
closing this as this is fixed in v2

PR #273, opened 2023-01-25
V2

thirtytwobits:
Merging v2 into main for release


thirtytwobits:
Looks like we're green on the CI finally. Just need a review of the merge.

pavel-kirienko (commented):
I did not actually review the code as the diff is large, but I understood this is not required in this case, as we all agree on where this is going.

pavel-kirienko (approved):

PR #276, opened 2023-01-29
Fixes #275, fixing python 3.11

### LINKED ISSUE #275: Python 3.11 support
3.11 breaks the work of Nunavut. Fix Nunavut.

thirtytwobits:


pavel-kirienko (approved):
Maybe it's time to upgrade to Ubuntu 22.04 while you're at it, at least for python 3.11?

thirtytwobits:
> Maybe it's time to upgrade to Ubuntu 22.04 while you're at it, at least for python 3.11?

We're using "latest" for the primary test runs. 20.04 is fixed for the backwards compatibility tests so we don't have what happened with python 3.6 again; where that old version was no longer available on ubuntu-latest.

PR #284, opened 2023-03-02
Add bool specialization for VariableLengthArray; fix AUTOSAR compliance issues; fix GTest integration

### LINKED ISSUE #260: Missing includes in c++ serialization.hpp
missing array and limits includes for C++ serialization. Also can remove climits and possibly other C includes that were carried over from the C

### LINKED ISSUE #281: C++: Missing bool specialization for the variable-length array
This is a critical issue because it makes the register API unusable with C++-generated types. For reference, in C this is supported via bitpacked arrays:

https://github.com/OpenCyphal/nunavut/blob/b3b6939dcf066e9b313bb120f2e47bb0d7bad73d/src/nunavut/lang/c/templates/definitions.j2#L139

pavel-kirienko:
- Add bool specialization for VariableLengthArray; fix #260 
- Fix #281 
- Update Google Test to the latest version v1.13
- Fix incorrect handling of compilation flags in `Findgtest.cmake`
- Fix warnings revealed by the above fix

There is a fair amount of code duplication between the generic and specialized versions of `VariableLengthArray`. It is possible to eliminate some of it by extracting commonalities into a base class.

### DIFF
--- a/src/nunavut/lang/cpp/support/variable_length_array.hpp
+++ b/src/nunavut/lang/cpp/support/variable_length_array.hpp
@@ -871,26 +871,23 @@ public:
     using size_type       = std::size_t;
     using allocator_type  = typename std::allocator_traits<Allocator>::template rebind_alloc<Storage>;
 
-private:
-    // The reference does not need to be copyable because URVO will elide the copy.
-    template <typename A>
-    class ReferenceImpl final
+    class reference final
     {
     public:
-        ReferenceImpl(const ReferenceImpl&) = default;
-        ReferenceImpl(ReferenceImpl&&)      = default;
-        ~ReferenceImpl()                    = default;
+        reference(const reference&) = default;
+        reference(reference&&)      = default;
+        ~reference()                = default;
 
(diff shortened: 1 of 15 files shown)


src/nunavut/lang/cpp/support/variable_length_array.hpp
```diff
@@ -849,16 +845,743 @@ class VariableLengthArray
...
+    };
+
+public:
+    using reference       = ReferenceImpl<VariableLengthArray>;
+    using const_reference = ReferenceImpl<const VariableLengthArray>;
+
+    using iterator       = IteratorImpl<VariableLengthArray>;
+    using const_iterator = IteratorImpl<const VariableLengthArray>;
+
+    VariableLengthArray() noexcept
```
thirtytwobits, line 1043: Should we make the default constructors for VariableLengthArray and this specialization, constexpr?

thirtytwobits (approved):
This is a lot of work. Thanks for contributing to Nunavut!

PR #287, opened 2023-03-03
C++: Avoid name collisions and add missing const begin()/end() methods to VLA

### LINKED ISSUE #286: C++: name collision between DSDL attributes and generated entities
C and Python implementations name all auxiliary entities following the reserved pattern `_.*_`, which mitigates the risk of name collision.

The C++ implementation mixes auxiliary and DSDL entities which is risky. It is not that difficult to run into a situation where a DSDL definition contains an item named, say, `serialize`.

pavel-kirienko:
Fixes #286 

In C++, `serialize` and `deserialize` are now free-standing functions. The alternative would be to name them `_serialize_` and `_deserialize_`, but I imagine this to be bad form.

One unused error tag `SERIALIZATION_INVALID_ARGUMENT` has been removed. The others have been renamed into CamelCase for consistency, as ALL_CAPS is reserved for macros.

I also added `_traits_::TypeOf` for introspection; it is often helpful in user code as the alternative requires the use of `decltype`:

image

We may also introduce a wrapper like `nunavut::traits<T>` if direct usage of `_traits_` is considered hard to read.

### DIFF
--- a/src/nunavut/lang/cpp/templates/_composite_type.j2
+++ b/src/nunavut/lang/cpp/templates/_composite_type.j2
@@ -33,42 +33,53 @@
 {% endif -%}
 {{ composite_type | definition_begin }} final
 {
-    // +---------------------------------------------------------------------------------------------------------------+
-    // | PORT IDENTIFIERS
-    // +---------------------------------------------------------------------------------------------------------------+
+    struct _traits_  // The name is surrounded with underscores to avoid collisions with DSDL attributes.
+    {
+        _traits_() = delete;
 {%- if T.has_fixed_port_id %}
-    static constexpr bool HasFixedPortID = true;
-    static constexpr {{ typename_unsigned_port }} FixedPortId = {{ T.fixed_port_id }}U;
+        static constexpr bool HasFixedPortID = true;
+        static constexpr {{ typename_unsigned_port }} FixedPortId = {{ T.fixed_port_id }}U;
 {%- else %}
-    /// This type does not have a fixed port-ID. See https://forum.opencyphal.org/t/choosing-message-and-service-ids/889
-    static constexpr bool HasFixedPortID = false;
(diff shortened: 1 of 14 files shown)


src/nunavut/lang/cpp/templates/_composite_type.j2
```diff
@@ -28,44 +28,58 @@
...
+            {{ composite_type.extent // 8 }}UL;
+        static constexpr {{ typename_unsigned_length }} SerializationBufferSizeBytes = {# -#}
+            {{ composite_type.inner_type.extent // 8 }}UL;
+        static_assert(ExtentBytes >= SerializationBufferSizeBytes, "Internal constraint violation");
+        static_assert({# -#}
+            ExtentBytes < (std::numeric_limits<{{ typename_unsigned_bit_length }}>::max() / 8U), {# -#}
+            "This message is too large to be handled by the selected types");
+{%- for field in composite_type.fields_except_padding %}
+    {%- if loop.first %}
+        struct TypeOf
```
thirtytwobits, line 73: TypeOf needs docs. Also, this is a really generic name that suggest some greater type system. Perhaps call it `TypeOfService` such that you get `TypeOfService::Response` which is a nice, English sentence "The type of the service response." etc.

pavel-kirienko:
We already have IndexOf so here I am trying to follow the same convention.
Do you think TypeOfAttribute is a better name? "Service" does not seem
related.

src/nunavut/lang/cpp/templates/_composite_type.j2
thirtytwobits: As long as you think this convention will be stable. We need to start closing the design of the C++ objects so we shouldn't toss in ergonomic things that we aren't confident are going to stand up over time.

thirtytwobits (approved):

Issue #288, opened 2023-03-03
The local testing insturctions are obsolete


pavel-kirienko:
Aren't we supposed to be using ghcr instead of dockerhub?

<img width="388" alt="image" src="https://user-images.githubusercontent.com/3298404/222776629-38d80434-c121-43bd-bec5-2defc50c96a1.png">

If it was possible to make these instructions less prone to obsolescence, that would be helpful.

thirtytwobits:
This should be fixed with 2.0.7

Issue #283, opened 2023-03-02
C++: Fixed length boolean arrays should be represented as std::bitset


pavel-kirienko:
cc @thirtytwobits @ASMfreaK 

This is critical because without this optimization, `uavcan.port.List` cannot be used in small embedded systems. See https://nunaweb.opencyphal.org/api/storage/docs/docs/uavcan/index.html#uavcan_node_port_SubjectIDList_0_1

pavel-kirienko:
Looking at the current implementation, I assume this will require an extension of the configuration parameter set unless we want to hard-code std::bitset for booleans.

thirtytwobits:
This should be fixed with 2.0.7

Issue #271, opened 2023-01-24
[C++ Headers] Non-sensical auto-generated if-statement leads to compilation error when compiled with `-Werror`


aentinger:
As the compiler says, a uint8_t can not have an larger value than 255, consequently the if statement is dead code.

**Compiler error**
```bash
reg/udral/service/battery/Status_0_2.hpp:462:26: error: comparison is always false due to limited range of data type [-Werror=type-limits]
  462 |             if ( _size0_ > 255U)
```

**reg/udral/service/battery/Status_0_2.hpp**
```C++
        const std::uint8_t _size0_ = in_buffer.getU8(8U);
        in_buffer.add_offset(8U);
            if ( _size0_ > 255U)
```

aentinger:
I'm only partially convinced. If OpenCyphal targets reliable systems its to be expected that whole codebase (including auto-generated code) does not break when using `-Werror`. This error may be trivial, but others may be not.

pavel-kirienko:
The matter of `-Werror` is orthogonal to the integrity of the codebase. Compilers are known to occasionally produce false diagnostics, which are promoted to build-breaking errors when strict compilation settings are given. This causes the compiler to reject valid code. It is, generally speaking, impossible to produce any non-trivial code that is certain to emit no diagnostics from all existing standard-compliant compilers. It follows that:

1. `-Werror` and similar strict options should only be used with the code you control directly and never be applied to third-party code.
2. Build error with `-Werror` does not indicate that the code is invalid or otherwise unacceptable.

thirtytwobits:
I'm inclined to accept this bug. I don't like normalizing deviance and if we can generate code that avoid well-known compiler warnings we should.

thirtytwobits:
This should be fixed with 2.0.7

PR #289, opened 2023-03-03
Fixes for #288, #283, and #271

thirtytwobits:
- #288 : Updating dev container documentation to point to ghcr.
- #283 : Detecting fixed boolean arrays and modifying c++ to emit std::bitset
- #271 : Changing the type used when comparing array lengths to avoid compiler warnings.


pavel-kirienko (approved):

pavel-kirienko:
The serialization calls need to be updated to use freestanding functions and also the version number needs updating.

thirtytwobits:
> The serialization calls need to be updated to use freestanding functions and also the version number needs updating.

Why did we do this?

pavel-kirienko:
>Why did we do this?

To avoid name clashing.

Issue #190, opened 2021-02-22
[C] Attach type version numbers to service types rather than nested request/response


Cherish-forever:
exp:
#define uavcan_node_ExecuteCommand_Request_1_1_EXTENT_BYTES_                    300UL
#define uavcan_node_ExecuteCommand_Request_1_1_SERIALIZATION_BUFFER_SIZE_BYTES_ 258UL

I think define like this will be better:
#define uavcan_node_ExecuteCommand_1_1_Request_EXTENT_BYTES_                    300UL
#define uavcan_node_ExecuteCommand_1_1_Request_SERIALIZATION_BUFFER_SIZE_BYTES_ 258UL

it will be easy to replace string "ExecuteCommand_1_1" for different version or different protocol.

pavel-kirienko:
Somewhat relevant discussion here: https://github.com/UAVCAN/nunavut/pull/155

Either format should work but I don't think it's worth changing it just for the sake of auto-renaming convenience.

thirtytwobits:
Unfortunately we've already discussed where the version should be and decided it belongs on the request and response types. This isn't something we want to make a major/breaking change for even if we could be convinced to change our previous decision. Sorry.

PR #290, opened 2023-03-04
C++: Fix incorrect handling of blank lines in doc comments; fix a minor mistake introduced in the prior PR

pavel-kirienko:
- Do not lose blank lines in doc comments.
- Move `Request`/`Response` aliases out of the `_traits_`; this was changed by mistake in #287
- Repair the vscode run/debug configuration

---

I find the naming of the service pseudo-type `Service_{{major}}_{{minor}}` non-obvious and I would like to suggest renaming it such that the service pseudo-type is named after the service itself suffixed with the version numbers. This would be in line with the way it is implemented in Python. Instead of this:

```c++
uavcan::_register::Access::Request_1_0
uavcan::_register::Access::Response_1_0
uavcan::_register::Access::Service_1_0
```

We would have this:

```c++
uavcan::_register::Access::Request_1_0
uavcan::_register::Access::Response_1_0
uavcan::_register::Access_1_0
```

Then there already are aliases inside `Access_1_0` which means that the full set of the available names is as follows:

```c++
uavcan::_register::Access::Request_1_0
uavcan::_register::Access::Response_1_0
uavcan::_register::Access_1_0
uavcan::_register::Access_1_0::Request
uavcan::_register::Access_1_0::Response
```

See related discussion in #155 and #190


ASMfreaK:
I'm sorry, if I've pulled the draft status too soon, I just wanted to say, that this is a great simplification! I like this a lot!

pavel-kirienko:
>I like this a lot!

Is this sarcasm or do you really support the proposal? Anyway, I think I am going to implement it now; shall Scott be against it, it can be reverted later.

ASMfreaK (approved):
My initial comment wasn't sarcasm. I liked the feature so much, that I came running to push the approve pr button and changed the draft status instead. I approve.

thirtytwobits (approved):

Issue #178, opened 2021-01-14
Default C language option "limit_empty_lines = 1" appears to be ineffective


pavel-kirienko:
```bash
$ pip install -U nunavut
$ nnvg --version
1.0.0%  
$ nnvg --target-language c --enable-serialization-asserts dsdl_src/my_project
```

Transpiled code contains more than 1 consecutive blank lines: [nunavut_out.tar.gz](https://github.com/UAVCAN/nunavut/files/5814125/nunavut_out.tar.gz)

aentinger:
I can confirm this by my own observation although it didn't happen on ever file. (https://github.com/107-systems/107-Arduino-UAVCAN/pull/65/commits/da38bc70ba48247c4022af8a4a72fadbae1295ee)

pavel-kirienko:
This is still valid as the root cause remains unaddressed.

thirtytwobits:
I'm removing all post-processing from Nunavut soon. Any whitespace bugs should be addressed by fixing the templates.

PR #291, opened 2023-03-05
Fix formatting inconsistencies in the C code generator (not C++)

pavel-kirienko:
- Closes https://github.com/OpenCyphal/nunavut/issues/178
- Also, fix benign warnings about uninitialized reads in C++ `VariableLengthArray<bool>`

### DIFF
--- a/src/nunavut/lang/cpp/support/variable_length_array.hpp
+++ b/src/nunavut/lang/cpp/support/variable_length_array.hpp
@@ -1413,11 +1413,15 @@ public:
 #endif
         if (new_data != nullptr)
         {
-            std::fill_n(new_data, bits2bytes(no_shrink_capacity), 0);  // Avoid uninitialized reads during modification
             if (new_data != data_)
             {
+                std::fill_n(new_data, bits2bytes(no_shrink_capacity), 0);
                 move_and_free(new_data, data_, size_, capacity_, alloc_);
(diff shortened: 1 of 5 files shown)


pavel-kirienko:
The variant support in C++ requires further work to add certain missing critical features. At first, I considered adding the required changes to this PR but then decided against it as they are comparatively significant, so there will be another changeset later.

src/nunavut/lang/cpp/support/variable_length_array.hpp
```diff
@@ -1413,6 +1413,7 @@ class VariableLengthArray<bool, MaxSize, Allocator>
 #endif
         if (new_data != nullptr)
         {
+            std::fill_n(new_data, bits2bytes(no_shrink_capacity), 0);  // Avoid uninitialized reads during modification
```
thirtytwobits, line 1416: I'm pretty sure this is wrong (I'm also surprised it doesn't break the tests). If you fill this pointer but the allocator had extended the existing memory then you just blew away the contents of the variant.

thirtytwobits (changes requested):

src/nunavut/lang/cpp/support/variable_length_array.hpp
pavel-kirienko: You are right. Fixed in the follow-up commit.

Issue #298, opened 2023-03-29
Bug when using std::vector in place of variable-length array type.


hungatamazon:
When following https://nunavut.readthedocs.io/en/latest/docs/languages.html to use `std::vector` instead of `variable_length_array`, there is a bug in the generated code.

The generated header has the intended type. Example: std::vector<std::uint32_t> value{}; . But it's trying to push_back() with no argument in the deserialize code.

```
    std::vector<std::uint32_t> value{};

...

    nunavut::support::SerializeResult
    deserialize(nunavut::support::const_bitspan in_buffer)
    {

     ...

            for (size_t _index5_ = 0U; _index5_ < _size0_; ++_index5_)
            {
                // TODO This is terribly inefficient. We need to completely refactor this template to use C++ emplace and
                // move semantics instead of assuming C-style containers
                value.push_back();   // <--- This causes compiler errors
                value[_index5_] = in_buffer.getU32(32U);
                in_buffer.add_offset(32U);
            }

      ...

    }
```

PR #303, opened 2023-05-17
Fix for #298 and workaround for Sybil 4 bug

thirtytwobits:
#298 will be fixed but this will also get another refactor when we switch to CETL.


thirtytwobits:
This PR is blocked by https://github.com/OpenCyphal/docker_toolchains/pull/34

verification/.devcontainer/devcontainer.json
```diff
@@ -12,13 +12,12 @@
...
-                "ban.spellright",
-                "ms-python.python",
-                "ms-vscode.cpptools",
                 "ms-vscode.cpptools-extension-pack",
+                "streetsidesoftware.code-spell-checker",
                 "xaver.clang-format",
-                "hbenl.vscode-test-explorer",
-                "matepek.vscode-catch2-test-adapter"
+                "vadimcn.vscode-lldb",
+                "matepek.vscode-catch2-test-adapter",
```
pavel-kirienko, line 19: What do we need catch2 support here for?

src/nunavut/lang/py/support/serialization.j2
pavel-kirienko, line 1: This one is clearly added by mistake.

src/nunavut/lang/js/support/serialization.j2
pavel-kirienko, line 1: This one is clearly added by mistake. If this is a placeholder, then it should be empty.

src/nunavut/lang/html/support/serialization.j2
pavel-kirienko, line 1: same

requirements.txt
```diff
@@ -1,6 +1,7 @@
 # This file provided for readthedocs.io only. Use tox.ini for all dependencies.
 
-sphinx >= 1.8.5
+# sphinx 7.0 is not compatible with sphinx_rtd_theme
+sphinx == 6.2.1
```
pavel-kirienko, line 4:
```suggestion
sphinx ~= 6.2.1
```

src/nunavut/cli/__init__.py
```diff
@@ -562,9 +562,7 @@ def main() -> int:
     #
     extra_includes = args.lookup_dir
 
-    # legacy variable. We'll support this for a time.
-    extra_includes_from_env = _extra_includes_from_env("UAVCAN_DSDL_INCLUDE_PATH")
-    extra_includes_from_env += _extra_includes_from_env("DSDL_INCLUDE_PATH")
+    extra_includes_from_env = _extra_includes_from_env("DSDL_INCLUDE_PATH")
```
pavel-kirienko, line 565: This might add unnecessary redundancy next to the `CYPHAL_PATH` environment variable that has become an almost de-facto standard already, as it is used in PyCyphal, Yakut, and Yukon and may be used by other implementations eventually. It contains the list of directories that contain DSDL root namespaces in them, separated with either `:` or `;`.

We may want to start using the same environment variable in Nunavut, such that the entries pulled through this variable are added after the entries given through the command line.

If we don't want to rely on the same env var in Nunavut, then at least `DSDL_INCLUDE_PATH` should be renamed to follow some consistent naming pattern, and the word "include" is better be removed since there is no concept of inclusion in DSDL.

pavel-kirienko (changes requested):

verification/.devcontainer/devcontainer.json
thirtytwobits: It's just a bad id. This support googletest:

```
Name: C++ TestMate
Id: matepek.vscode-catch2-test-adapter
Description: Run GoogleTest, Catch2 and DOCtest tests from VSCode
Version: 4.4.2
Publisher: Mate Pek
VS Marketplace Link: https://marketplace.visualstudio.com/items?itemName=matepek.vscode-catch2-test-adapter
```

src/nunavut/lang/py/support/serialization.j2
thirtytwobits: fixed

src/nunavut/cli/__init__.py
thirtytwobits: I'll just support them all.

thirtytwobits:
For the love of Pete! Is the entire Python universe broken!?! (working on a fix for RTD)

pavel-kirienko (approved):
Not sure if I get it correctly but `CYPHAL_PATH` contains the list of directories that contain DSDL root namespaces in them. That's how it's defined in PyCyphal et al. Is this compatible with the semantics of your env var here?

thirtytwobits:
> It seems like Nunavut expects the env vars to contain the list of root namespace directories,...

Undoing then. Can you open an issue to add support for CYPHAL_PATH?

skeetsaz (approved):

PR #316, opened 2023-08-08
Add support for CETL VariableLengthArray and polymorphic_allocator

skeetsaz:
This change modifies the C++ code-gen to create message classes that can be used with allocators.
- The message class becomes an "_Impl" template that takes an allocator, which gets passed down to any variable length arrays in the messages (or in contained messages.)  The normal message name is then aliased to the "_Impl" template using the allocator specified in the yaml config.  In the end the message is a regular struct as it was before.
- The old variable_length_array.hpp is removed from the support library.  CETL is added as a submodule and the new variable_length_array.hpp from there is used.
- Added a verification test to exercise a message using CETL's polymorphic_allocator


src/nunavut/lang/cpp/templates/_composite_type.j2
```diff
@@ -31,8 +31,69 @@
...
+    // Copy assignment
+    {{composite_type|short_reference_name}}_Impl& operator=(const {{composite_type|short_reference_name}}_Impl& rhs)
+    {
+    {%- for field in composite_type.fields_except_padding %}
+        {{ field | id }} = rhs.{{ field | id }};
+    {%- endfor %}
+        return *this;
+    }
+
+    // Move assignment
```
skeetsaz, line 88: For the special methods that don't take an allocator I suppose "= default" would supply the same implementation, right?  If that's correct I could shorten the code-gen for those cases.

thirtytwobits:
You need to run Black to get the CI past the formatting errors.

src/nunavut/lang/cpp/__init__.py
```diff
@@ -968,15 +949,80 @@ def filter_destructor_name(language: Language, instance: pydsdl.Any) -> str:
 
 
 @template_language_filter(__name__)
-def filter_default_value_initializer(language: Language, instance: pydsdl.Any) -> str:
+def filter_value_initializer(language: Language, instance: pydsdl.Any, initialization_context: SpecialMethods) -> str:
```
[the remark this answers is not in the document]
thirtytwobits: @pavel-kirienko, what do you think of my CLI proposal here?

thirtytwobits (changes requested):

src/nunavut/lang/cpp/__init__.py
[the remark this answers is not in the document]
pavel-kirienko: lgtm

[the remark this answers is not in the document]
skeetsaz: This sounds like a good plan though it will take me some time to sort out the best way to implement it.  Doing a quick search I don't see much logic hinging on the specified language standard -- is there not much use of that currently?  Do you think should try to keep to a single _composite_type.j2 template or break it out into multiples?  Let me know if you have any suggestions on a general strategy otherwise I'll dive in and figure something out.

src/nunavut/lang/cpp/__init__.py
```diff
@@ -968,15 +949,80 @@ def filter_destructor_name(language: Language, instance: pydsdl.Any) -> str:
...
     return ""
 
 
+@template_language_filter(__name__)
+def filter_field_declaration(language: Language, instance: pydsdl.Any) -> str:
+    """
+    Emit a field declaration that accounts for the message being a templated type
+    """
+    suffix: str = ""
+    if isinstance(instance, pydsdl.CompositeType):
```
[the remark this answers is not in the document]
skeetsaz: Yeah, I was trying to do that with the using alias but your suggested approach above will let me more cleanly generate to the same non-breaking type name.

skeetsaz:
Updated my pull-request per the feedback.  The ctor_convention approach and -std=cetl++ -std=c++17-pmr flags help simplify things and maintain better backward compatibility.

The verification tests I added will fail due to CETL VLA not implementing the trailing-allocator convention correctly (I filed https://github.com/OpenCyphal/CETL/issues/57 for that.)  But I want to get this out for more feedback.

docs/languages.rst
```diff
@@ -16,25 +16,50 @@ See :ref:`template-language-guide` until this section is more complete.
...
+      options:
+        variable_array_type_include: "<vector>"
+        variable_array_type_template: "std::vector<{TYPE}, {REBIND_ALLOCATOR}>"
+        variable_array_type_constructor_args: ""
+        allocator_include: "<memory>"
+        allocator_type: "std::pmr::polymorphic_allocator"
+        allocator_is_default_constructible: true
+        ctor_convention: "uses-trailing-allocator"
+
+cetl++.yaml
```
thirtytwobits, line 42: Please add to this document that `cetl++ == c++14` but with extra stuff from CETL.

src/nunavut/lang/__init__.py
```diff
@@ -220,6 +222,13 @@ def set_target_language(self, target_language: typing.Optional[str]) -> "Languag
             self._target_language_name = LanguageClassLoader.to_language_name(target_language)
         return self
 
+    def load_default_config(self, language_standard: str) -> None:
+        self._ln_loader.config  # Accessing this property causes the defaults to load
+        defaults_key = f"{language_standard}_options"
+        if defaults_key in self._ln_loader.config.sections()[NUNAVUT_LANG_CPP]:
```
thirtytwobits, line 228: This is a leaky abstraction. `nunavut.lang` should have no specific knowledge of `nunavut.lang.cpp`

src/nunavut/lang/cpp/__init__.py
```diff
@@ -62,6 +82,19 @@ def _handle_stropping_or_encoding_failure(
         # we couldn't help after all. raise the pending error.
         raise pending_error
 
+    def _validate_language_options(self, language_options: typing.Mapping[str, typing.Any]) -> None:
```
thirtytwobits, line 85: This functionality belongs in `nunavut.lang.LanguageConfig`. While it's a bad design to code specific configuration rules into the generic LanguageConfig class it is a good stop-gap until we can incorporate yaml schemas of some kind. By putting these warts inside the LanguageConfig abstraction we can minimize the impact of adding schema validation the reset of the code.

src/nunavut/lang/cpp/__init__.py
```diff
@@ -977,6 +966,77 @@ def filter_default_value_initializer(language: Language, instance: pydsdl.Any) -
...
+
+        if wrap:
+            rhs = "{}({})".format(wrap, rhs)
+        args = []
+        if rhs:
+            args.append(rhs)
+        args = leading_args + args + trailing_args
+        return "{" + ", ".join(args) + "}"
+
+    return ""
```
thirtytwobits, line 1037: nit, but this method would be a touch nicer if we had a single return value:
```
filtered_value: str = ""
if (
...
    filtered_value = "{" + ", ".join(args) + "}"

return filtered_value
```

src/nunavut/lang/cpp/templates/_composite_type.j2
```diff
@@ -33,6 +33,52 @@
 {% endif -%}
 struct {{composite_type|short_reference_name}} final
 {
+{%- if options.ctor_convention != ConstructorConvention.Default.value %}
+    using allocator_type = {{ options.allocator_type }}<decltype(nullptr)>;
```
thirtytwobits, line 37: This belongs in the `_traits_` struct to avoid name collisions.

thirtytwobits (changes requested):
Is there a way to specify the default memory resource that I'm missing?

thirtytwobits (approved):

PR #319, opened 2023-10-11
Add field initializing constructor, fix explicit, fix deserialization bug

skeetsaz:
When you use "cetl++14-17" or "c++17-pmr" we code-gen message classes with
constructors that take an allocator.  Since the class has a user-defined
constructor this means it can no longer use aggregate initialization
(https://en.cppreference.com/w/cpp/language/aggregate_initialization)
This is inconvenient so I'm adding a constructor with args for each field,
in order.  This also makes it possible to declare a const message instance.

Some single arg constructors weren't marked explicit, thus becoming user-defined
conversion functions, which introduces unexpected bugs. I've fixed all the
constructors so any single-arg ones are declared explicit.

Fixed a deserialization bug where the allocator was not getting passed to a
temporary.


skeetsaz:
This is showing old commits so I think I put up the pull request incorrectly.  Will put up a new one.

PR #323, opened 2023-11-18
Fix issues: allocator passing, verification test C++ flavors, union code-gen

skeetsaz:
- Changed the verification tests CMakeLists.txt so it uses a matrix of test files to language flavors.  This allows turning language flavors on or off for each test.
- Updated the docker image tag used by the verification tests.
- Fixed the c++17-pmr flavor to use "memory_resource" as the allocator include since that's where std::pmr::polymorphic_allocator is.
- Exclude unions from all code-gen using allocator since there is not a way to pass it through to the elements of std::variant in a way that makes sense.


pavel-kirienko:
>Exclude unions from all code-gen using allocator since there is not a way to pass it through to the elements of std::variant in a way that makes sense.

Could you please expand a bit on the implications of this for user code? Will one still be able to use unions with pmr?

skeetsaz:
I looked into using unions with PMR and it looks like it is indeed do-able.  I worked out what the code-gen should look like here.  There are a few subtleties but I think it's a good way to go.
https://godbolt.org/z/qvWTnhEd8

If you don't mind, I'd like to move forward with what I have here and do the union work in a separate PR.  The current code checked into github is broken (not sure why the github automated tests aren't running everything but if you pull it down locally and run all the verification tests they are broken.)  This will get them all compiling/passing again.

Thanks for bringing this up.  I was not too familiar with std::variant so it forced me to read up on it and understand it better.

skeetsaz:
All the tests pass except for readthedocs which seems to be broken due to some infrastructure issue.

thirtytwobits (approved):

PR #326, opened 2024-01-18
Allow setting support_templates_dir additionally to templates_dir

swuerl:
Using an external template for support code is currently not possible, as it was explicitly disabled to fix a bug:
https://github.com/OpenCyphal/nunavut/commit/c7db1bd6bf626585eaa4e1355f6bd82d99b48320

I was working around that bug in the past, by just calling nunavut twice, for support and templates, e.g.:

```
nnvg --templates my_template/support --target-language c --generate-support never ...
nnvg --templates my_template/templates --target-language c --generate-support only ...
```

Since version 2.1.0, this is not possible anymore. This PR implements this use-case properly again, by adding an argument `--support-templates` to separate support from struct serialization. So now it becomes:

```
nnvg --support-templates my_template/support  --templates my_template/templates --target-language c ...
```


thirtytwobits:
Sorry for the delay on this. I'm looking at the PR now...

thirtytwobits:
This is great. I'll open an issue to refactor that constructor as it's gotten too complicated and should take a configuration object instead. Thanks @swuerl

thirtytwobits (approved):

PR #333, opened 2024-03-16
[python] Make the Python language verification suite pass against Python 3.12

### LINKED ISSUE #332: Add Python 3.12 support for generated python
The current nunavut_support.py.jinja template produces type annotations that are not compatible with both python 3.11 and 3.12. Fix and add 3.12 test to the noxfile

pavel-kirienko:
Fix #332


pavel-kirienko:
@thirtytwobits it works fine on my local machine with Python 3.12.2, but fails in CI with an obscure internal error originating from the `coverage` package. There are a couple of odd reports related to it:

- https://github.com/nedbat/coveragepy/issues/1634
- https://github.com/nedbat/coveragepy/issues/1376

I suspect at the moment that this is not our problem; the latest CI run will reveal more info.

pavel-kirienko:
Hah! We're running an alpha build of Python: https://github.com/OpenCyphal/nunavut/actions/runs/8310952438/job/22744051997?pr=333#step:4:269

@thirtytwobits we're going to need a newer Python, our code is fine

thirtytwobits (approved):

PR #343, opened 2024-08-27
Add `FullNameAndVersion` to c++

serges147:


serges147:
> Can you change tox.ini to remove `types-pkg_resources` and add `types-setuptools`? ...

@thirtytwobits I did, now it fails on doctest

PR #349, opened 2024-10-14
Add support of "c++20-pmr" c++ language flavor

serges147:
- Needed for libcyphal project where I'm planning to use it like this:
```
    if (CMAKE_CXX_STANDARD STREQUAL "14")
        list(APPEND LOCAL_NNVG_CMD_ARGS "--language-standard=cetl++14-17")
    else ()
        list(APPEND LOCAL_NNVG_CMD_ARGS "--language-standard=c++${CMAKE_CXX_STANDARD}-pmr")
    endif ()
```
- `cetl++14-17` now uses `cetl::pf17::variant` for union types.
- "pmr" language feature now propagates internal `allocator_` field.
- Generation of a type members initialization a bit changed (stylistically) for easier generation (`:` now on a new line)

BTW, these changes are successfully in use at libcyphal PR: https://github.com/OpenCyphal-Garage/libcyphal/pull/389
where I bring up registry list/access servers at Example_2_Application_0_NodeHeartbeatGetInfo_Udp .

### DIFF
--- a/src/nunavut/lang/cpp/support/serialization.j2
+++ b/src/nunavut/lang/cpp/support/serialization.j2
@@ -931,12 +931,12 @@ inline {{typename_float_64}} const_bitspan::getF64()
 }
 
 template <typename T>
-void copyAssignInplace(T& dst, const T& src)
+void assignInplace(T& dst, T&& src)
 {
     // Intentional violation of MISRA cpp:M23_329: "Advanced memory management shall not be used".
     // This method is used to copy-assign objects in-place (like PMR allocators). It is safe and efficient.
     dst.~T();  // NOLINT NOSONAR
-    new (&dst) T(src);
+    new (&dst) T(std::forward<T>(src));
 }
 
 {% endif -%}
--- a/src/nunavut/lang/cpp/templates/_composite_type.j2
+++ b/src/nunavut/lang/cpp/templates/_composite_type.j2
@@ -114,7 +114,7 @@ struct {% if composite_type.deprecated -%}
     {%- if composite_type.inner_type is UnionType %}
         : union_value{}
         {%- if not composite_type.bit_length_set.fixed_length %}
-        , allocator_{allocator}
+        , allocator_{allocator.select_on_container_copy_construction()}
         {%- endif %}
     {%- else %}
     {%- for field in composite_type.fields_except_padding %}
@@ -153,7 +153,7 @@ struct {% if composite_type.deprecated -%}
     {%- if composite_type.inner_type is UnionType %}
         : union_value{rhs.union_value}
         {%- if not composite_type.bit_length_set.fixed_length %}
-        , allocator_{allocator}
+        , allocator_{allocator.select_on_container_copy_construction()}
         {%- endif %}
     {%- else %}
     {%- for field in composite_type.fields_except_padding %}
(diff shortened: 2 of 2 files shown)


src/nunavut/lang/cpp/support/serialization.j2
```diff
@@ -930,6 +930,15 @@ inline {{typename_float_64}} const_bitspan::getF64()
     return tmp.fl;
 }
 
+template <typename T>
+void copyAssignInplace(T& dst, const T& src)
```
thirtytwobits, line 934: This is incorrect. Please use `select_on_container_copy_construction` (https://en.cppreference.com/w/cpp/memory/polymorphic_allocator/select_on_container_copy_construction). See `variable_length_array` for example usage.

thirtytwobits (changes requested):

src/nunavut/lang/cpp/support/serialization.j2
serges147: @thirtytwobits Scott, is this what you are looking for?
https://github.com/OpenCyphal/nunavut/pull/349/commits/23b84f76d4fc0e75595efc02fde30140e3cef2fb
Honestly, I don't understand this part of c++ at all, and never encountered any book really describing such advanced topic of STL containers and their allocator traits and requirements.

thirtytwobits: If you look at `variable_length_array` you'll see it does this:

```
: alloc_{std::allocator_traits<allocator_type>::select_on_container_copy_construction(rhs_alloc)}
```

thirtytwobits: Additionally, in the copy assignment operator it adheres to the `propagate_on_container_copy_assignment` protocol defined by [allocator_traits](https://en.cppreference.com/w/cpp/memory/allocator_traits)

```
    // +----------------------------------------------------------------------+
    // | Allocator copy assignment
    // +----------------------------------------------------------------------+
    template <typename UAlloc>
    constexpr bool copy_assign_alloc(
        const UAlloc& rhs,
        typename std::enable_if_t<std::allocator_traits<UAlloc>::propagate_on_container_copy_assignment::value>* =
            nullptr) noexcept
    {
        static_assert(std::is_nothrow_copy_assignable<allocator_type>::value,
                      "The C++ standard requires anything adhering to propagate_on_container_copy_assignment to copy "
                      "without throwing.");
        alloc_ = rhs;
        return true;
    }

    template <typename UAlloc>
    constexpr bool copy_assign_alloc(
        const UAlloc& rhs,
        typename std::enable_if_t<!std::allocator_traits<UAlloc>::propagate_on_container_copy_assignment::value>* =
            nullptr) noexcept
    {
        (void) rhs;
        return false;
    }
```

Generally, the aliasing of memory here to do an in-place new on top of the current object seems dangerous. At the very least we'd need to understand if `std::launder` was needed in this case or not?

serges147: Scott, I don't understand why in general I need to worry about all these `select_on_container...` and `propagate_on_container` stuff... All I need is along a variant (for union nunavut types) also store `allocator_`. This allocator is needed in case when value of the variant will be reassigned to some other variable size value (and so I need allocator to be able pass as the last parameter of constructor).

serges147: >in-place new on top of the current object seems dangerous
I believe this is very close to what `optional::emplace` does

thirtytwobits: ok

PR #350, opened 2024-10-18
Reverted usage of select_on_container_copy_construction()

serges147:
Reverted usage of `select_on_container_copy_construction()` - it broke libcyphal unit tests where there are unit tests which verify that there is no usage of the default PMR.

`select_on_container_copy_construction()` does exactly this - it create a completely new instance of allocator:
- without transferring underlining memory resource
- but instead getting the default `pmr::memory_resource()`


serges147:
Closing this PR b/c with https://github.com/OpenCyphal-Garage/libcyphal/pull/394 it is not needed anymore.

Issue #348, opened 2024-10-09
C++ `support/serialization.hpp` should `#include <cassert>`


serges147:
Compilation fails if prior to `#include <nunavut/support/serialization.hpp>` there was no `assert` definition - as a result, client code is required to have it like this:
```
...
#include <cassert>  // NOLINT for NUNAVUT_ASSERT
#include <nunavut/support/serialization.hpp
...
// other c++ system headers
```

serges147:
Fixed by PR #352

PR #352, opened 2024-11-05
Expose `FullNameAndVersion()` from a DSDL type.

serges147:
1. Exposed full name of a DSDL type as...
```
struct _traits_
{
    ...
    constexpr const char* FullNameAndVersion() { return "{{ composite_type }}"; }
};
```
Needed for libcyphal's subscriber implementation (see [issue # 380](https://github.com/OpenCyphal-Garage/libcyphal/issues/380))

2. Fix for issue #348 - added `#include <cassert>` for `nunavut/support/serialization.hpp` file generation

### DIFF
--- a/src/nunavut/lang/cpp/templates/_composite_type.j2
+++ b/src/nunavut/lang/cpp/templates/_composite_type.j2
@@ -61,10 +61,11 @@ struct {% if composite_type.deprecated -%}
         static constexpr bool IsService      = false;
         static constexpr bool IsRequest      = {{ (composite_type == T.request_type) | string | lower }};
         static constexpr bool IsResponse     = {{ (composite_type == T.response_type) | string | lower }};
-{%- else %}
+{% else %}
         static constexpr bool IsServiceType = false;
-{% endif -%}
-    {%- assert composite_type.extent % 8 == 0 %}
+{% endif %}
+        static constexpr char FullNameAndVersion[] = "{{ composite_type }}";
+    {%  assert composite_type.extent % 8 == 0 %}
     {%- assert composite_type.inner_type.extent % 8 == 0 %}
         /// Extent is the minimum amount of memory required to hold any serialized representation of any compatible
         /// version of the data type; or, on other words, it is the the maximum possible size of received objects of this type.
--- a/src/nunavut/lang/cpp/templates/_composite_type.j2
+++ b/src/nunavut/lang/cpp/templates/_composite_type.j2
@@ -64,7 +64,7 @@ struct {% if composite_type.deprecated -%}
 {% else %}
         static constexpr bool IsServiceType = false;
 {% endif %}
-        static constexpr char FullNameAndVersion[] = "{{ composite_type }}";
+        static constexpr const char* FullNameAndVersion() { return "{{ composite_type }}"; }
     {%  assert composite_type.extent % 8 == 0 %}
     {%- assert composite_type.inner_type.extent % 8 == 0 %}
         /// Extent is the minimum amount of memory required to hold any serialized representation of any compatible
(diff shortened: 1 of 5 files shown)


src/nunavut/lang/cpp/templates/_composite_type.j2
```diff
@@ -61,10 +61,11 @@ struct {% if composite_type.deprecated -%}
         static constexpr bool IsService      = false;
         static constexpr bool IsRequest      = {{ (composite_type == T.request_type) | string | lower }};
         static constexpr bool IsResponse     = {{ (composite_type == T.response_type) | string | lower }};
-{%- else %}
+{% else %}
         static constexpr bool IsServiceType = false;
-{% endif -%}
-    {%- assert composite_type.extent % 8 == 0 %}
+{% endif %}
+        static constexpr const char* FullNameAndVersion() { return "{{ composite_type }}"; }
```
serges147, line 67: Unfortunately I can't make it as simple as could be in c++17:
`static constexpr char FullNameAndVersion[] = "{{ composite_type }}";`

pavel-kirienko (approved):

verification/cpp/suite/test_constant.cpp
```diff
@@ -0,0 +1,117 @@
+/*
+ * Copyright (c) 2024 OpenCyphal Development Team.
+ * Authors: Sergei Shirokov <sergei.shirokov@zubax.com>
```
serges147, line 3: Inspired by corresponding "C" test, namely `verification/c/suite/test_constant.c`

verification/cpp/suite/test_constant.cpp
```diff
@@ -0,0 +1,117 @@
...
+    EXPECT_THAT(regulated::basics::Struct__0_1::_traits_::FullNameAndVersion(), StrEq("regulated.basics.Struct_.0.1"));
+
+    // Application constants.
+    EXPECT_THAT(regulated::basics::Struct__0_1::CONSTANT_MINUS_THREE, DoubleNear(-3.0, 1e-9));
+    EXPECT_THAT(regulated::basics::Struct__0_1::CONSTANT_ZEE, 'Z');
+    EXPECT_THAT(-regulated::basics::Struct__0_1::CONSTANT_MINUS_MAX_OFFSET,
+                Le(regulated::basics::Struct__0_1::_traits_::SerializationBufferSizeBytes * 8U));
+    EXPECT_TRUE(regulated::basics::Struct__0_1::CONSTANT_TRUTH);
+
+// TODO: Uncomment when some form of "ARRAY_CAPACITY" is generated.
```
serges147, line 33: Separate issue and PR but probably I will add generation of a new `<DSDL Type>::_traits_::ArrayCapacity` struct,
where there will be listed `static constexpr std::size_t` constants for each field which has variable array type. Static array fields don't require such information (b/c `std::array<>` already has `.size()` method), but maybe just for consistency I will list them as well.

serges147: @thirtytwobits Do you agree with my future plan about above "ARRAY_CAPACITY" topic?

thirtytwobits: CETL's variable_length_array already provides max_size for this?

thirtytwobits (changes requested):

verification/cpp/suite/test_constant.cpp
serges147: Yes it does, but under c++17 and above we use `std::vector`

src/nunavut/lang/cpp/templates/_composite_type.j2
serges147: Sonar will catch ODR issue with such approach.

Issue #370, opened 2025-01-13
C++ generator: Incorrect de-serialization of delimited types.


serges147:
During de-serialization of a nested (not top level) delimited type (with `@extent`) its header size is not taken into account but instead whole remaining part of the buffer is in use - as a result "garbage" bytes could end up in the result deserialized instance when input buffer is longer than expected. `C` generator does it correctly.

serges147:
Fixed by PR #371

PR #371, opened 2025-01-13
Fix for issue 370 : C++ generator deserialization

serges147:
See issue #370

### DIFF
--- a/src/nunavut/lang/cpp/templates/deserialization.j2
+++ b/src/nunavut/lang/cpp/templates/deserialization.j2
@@ -222,7 +222,7 @@
 {% set ref_size_bytes = 'size_bytes' |to_template_unique_name %}
 {% set ref_delimiter  = 'dh' |to_template_unique_name %}
     {
-        {{ typename_unsigned_length }} {{ ref_size_bytes }};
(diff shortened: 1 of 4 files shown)


thirtytwobits (approved):

src/nunavut/lang/cpp/templates/deserialization.j2
```diff
@@ -222,20 +222,26 @@
 {% set ref_size_bytes = 'size_bytes' |to_template_unique_name %}
 {% set ref_delimiter  = 'dh' |to_template_unique_name %}
     {
-        {{ typename_unsigned_length }} {{ ref_size_bytes }} = in_buffer.size() / 8U;
+        {{ typename_unsigned_length }} {{ ref_size_bytes }};
```
pavel-kirienko, line 225: A default initializer is needed anyway. The old construct seems fine; the optimizer will throw the extra assignment away anyway, then you can remove the else branch at the bottom.

pavel-kirienko (approved):

src/nunavut/lang/cpp/templates/deserialization.j2
serges147: fixed

PR #372, opened 2025-02-14
rpi32: raspean bookworm has cmake 3.25 only

serges147:


pavel-kirienko (approved):

serges147:
@thirtytwobits Could you please take a look why CI is red? I believe it's not b/c of my change...

thirtytwobits:
Closing this as we do not need to build on raspean.

